PR #70, opened 2022-01-18
Deployment using Github OIDC provider [PRO-189]

kkowalski:
## Motivation

Fix our deploy jobs

## Solution

Use Github OIDC provider to authenticate to AWS

## PR Checklist

- [x] verify successful deploy

### DIFF
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -14,7 +14,7 @@ jobs:
   accept:
     name: Accept
     runs-on: ubuntu-latest
-    needs: [lint, test, build_and_push, image_manifest, helm_lint, helm_test]
+    needs: [lint, test, build_and_push, image_manifest, snyk_scan_container, helm_lint, helm_test]
     steps:
       - name: Accept
         run: true
@@ -288,7 +288,7 @@ jobs:
     name: Deploy main to staging
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main'
-    needs: [accept, image_manifest]
+    needs: [accept]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -320,7 +320,7 @@ jobs:
   deploy-stage-manual:
     name: Manual deploy to staging
     runs-on: ubuntu-latest
-    needs: [accept, image_manifest]
+    needs: [image_manifest]
     environment: stage-manual
     permissions:
       id-token: write
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -284,8 +284,8 @@ jobs:
           | cut -f 1 -d ' '
           | xargs -L 1 kubectl logs
 
-  deploy-stage:
-    name: Deploy to staging
+  deploy-main-stage:
+    name: Deploy main to staging
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test-deploy.yml
```diff
@@ -270,10 +270,8 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID_NEW }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY_NEW }}
```
lucasege, line 274: nit: can we use a better naming convention for these keys instead of `_NEW`? Just so we can support potential future upgrades without devolving into `_NEW_NEW...`.
Perhaps `_OIDC` or a version number?

lucasege (commented):
I'm not in our AWS so I can't see the `IAM` role configs to verify that those permissions are set correctly. If those are OK, I would say LGTM

.github/workflows/build-test-deploy.yml
kkowalski: After this change we won't store long term AWS credentials.
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers#overview

.github/workflows/build-test-deploy.yml
```diff
@@ -307,10 +309,8 @@ jobs:
...
           aws-region: ${{ env.AWS_REGION }}
-          role-to-assume: arn:aws:iam::267301337261:role/cross-account/deploy
+          role-to-assume: arn:aws:iam::267301337261:role/oidc/github-deploy
           role-duration-seconds: 900
       - run: aws eks --region us-east-1 update-kubeconfig --name $CLUSTER_NAME
       - name: Create pull secret for ghcr.io
```
recmo, line 316: Note to self: This should not be necessary now that the repo is public

.github/workflows/build-test-deploy.yml
```diff
@@ -298,6 +296,10 @@ jobs:
     runs-on: ubuntu-latest
     needs: [ test, build_and_push ]
     environment: stage-manual
+    permissions:
+      id-token: write
+      contents: read
+      packages: read
```
recmo, line 302: Reading up on this. What are the default permissions we use? Can we set them to 'restricted' ?

.github/workflows/build-test-deploy.yml
```diff
@@ -270,10 +270,8 @@ jobs:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID_NEW }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY_NEW }}
           aws-region: ${{ env.AWS_REGION }}
-          role-to-assume: arn:aws:iam::267301337261:role/cross-account/deploy
+          role-to-assume: arn:aws:iam::267301337261:role/oidc/github-deploy
           role-duration-seconds: 900
       - run: aws eks --region us-east-1 update-kubeconfig --name $CLUSTER_NAME
```
recmo, line 276: `us-east-1` -> `$AWS_REGION`

PR #92, opened 2022-03-17
Change domain in ingress

kkowalski:
## Motivation

"crypto" subdomain wasn't the best. We're routing based on domains not paths thus updating to:
`signup.stage-crypto.worldcoin.dev`, other services will be available at `hubble.stage-crypto.worldcoin.dev` and `mint.stage-crypto.worldcoin.dev`

## Solution

Updating ingress


kkowalski:
Closing this one in favor of #93

PR #121, opened 2022-04-28
fix event and remove tx check

philsippl:
## Motivation

- Fixes a bug in during syncing, in which roots were inserted instead of leaves (identityCommitments).
- Removes "checking" tx that tried to call `create_group` without a signer

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


m1guelpf (approved):
LGTM

PR #137, opened 2022-05-03
Overpay gas to ensure inclusion

philsippl:
## Motivation

We have no tracking of transactions implemented, if one tx gets dropped from mempool, we leave all pending txs hanging forever. This is a quick hack to ensure we overpay gas to reduce the risk of this happening.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -17,7 +17,7 @@ use std::{
     sync::atomic::{AtomicUsize, Ordering},
 };
 use structopt::StructOpt;
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::RwLock;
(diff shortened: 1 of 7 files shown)


src/app.rs
```diff
@@ -127,13 +128,22 @@ impl App {
             next_leaf = max(next_leaf, leaf + 1);
         }
 
+        let is_manager = ethereum.is_manager().await?;
```
recmo, line 131: Note that it can change over time.

recmo (approved):
Let's merge, but refactor later.

PR #149, opened 2022-05-11
Robuster eth stack

recmo:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/ethereum/mod.rs
```diff
@@ -45,252 +67,274 @@ pub struct Options {
...
 
-    #[structopt(
-        short,
-        parse(try_from_str),
-        default_value = "false",
-        env = "SIGNUP_SEQUENCER_MOCK"
-    )]
-    pub mock: bool,
+    /// Maximum number of blocks to pull events from in one request.
+    #[structopt(long, env, default_value = "1000")]
```
philsippl, line 71: let's replace `BLOCK_QUERY_RANGE` by `MAX_LOG_BLOCKS` in `.github/workflows/build-test-deploy.yml`

src/ethereum/mod.rs
```diff
@@ -45,252 +67,274 @@ pub struct Options {
     )]
     pub eip1559: bool,
 
-    #[structopt(
-        short,
-        parse(try_from_str),
-        default_value = "false",
-        env = "SIGNUP_SEQUENCER_MOCK"
-    )]
-    pub mock: bool,
```
philsippl, line 54: should be also removed from `.github/workflows/build-test-deploy.yml`

src/app.rs
```diff
@@ -125,88 +114,130 @@ impl App {
...
-        {
-            let mut merkle_tree = self.merkle_tree.write().await;
-            merkle_tree.set(identity_index, *commitment);
+        tree.set(identity_index, *commitment);
+
+        // Downgrade write lock to read lock
+        let tree = tree.downgrade();
+
+        // Check tree root
+        if let Err(error) = self.contracts.assert_valid_root(tree.root()).await {
```
philsippl, line 213: this will also fail if the root is expired: https://github.com/worldcoin/world-id-example-airdrop/blob/03de53d2cb016ddef28b26e8237e85b62ec385c7/src/Semaphore.sol#L143

then you can never insert an element again. We should prob introduce a separate error for that in the contracts. I would just drop the `assert_valid_root` check in this get handler to be backwards compatible with our existing deployment. Also, it feels not super necessary.

src/contracts/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    #[instrument(level = "debug", skip_all)]
+    pub async fn insert_identity(&self, commitment: &Field) -> EyreResult<()> {
+        info!(%commitment, "Inserting identity in contract");
+
+        // Send create tx
+        let commitment = U256::from(commitment.to_be_bytes());
+        self.ethereum
+            .send_transaction(self.semaphore.add_member(self.group_id, commitment).tx)
```
philsippl, line 189: We probably need to specify an explicit gas limit, I think the estimate fails, which leads to following error:
```
ERROR signup_sequencer::ethereum: Transaction failed, e: Send(MiddlewareError(MiddlewareError(MiddlewareError(MiddlewareError(JsonRpcClientError(JsonRpcError(JsonRpcError { code: -32000, message: "gas required exceeds allowance (5)", data: None })))))))
```

src/app.rs
recmo: The root should not have expired because it just got updated on line 204. This is the new root just after insertion.

That being said, the root _before_ insertion should also be valid because it should be the latest root. I think the contract might be wrong here.

src/contracts/mod.rs
recmo: Let's investigate this more than. If it's more than the +10% + 10k (which it uses now), then I'd like to know the actual numbers and know where the delta is coming from.

src/app.rs
philsippl: oh right, actually not an issue. I saw the error originally on the `/inclusionProof` endpoint. No issue here for insertion.

src/app.rs
```diff
@@ -217,24 +248,46 @@ impl App {
...
+                ?identity_commitment,
+                ?identity_index,
+                ?root,
+                "Proof does not verify locally."
+            );
+            panic!("Proof does not verify locally.");
+        }
+
+        // Verify the root on chain
+        if let Err(error) = self.contracts.assert_valid_root(root).await {
```
philsippl, line 269: Asking for inclusion proof for an outdated root (see above comment), panics the whole sequencer.

src/app.rs
```diff
@@ -217,24 +248,46 @@ impl App {
...
+            error!(
+                computed_root = ?root,
+                ?error,
+                "Root mismatch between tree and contract."
+            );
+            panic!("Root mismatch between tree and contract.");
```
philsippl, line 275: Asking for inclusion proof for an outdated root (see above comment), panics the whole sequencer.

PR #174, opened 2022-06-03
Batch insert endpoint

corydickson:
Starts the process to interact with the contracts using the new batch insertion method


recmo (commented):
Looks out of date wrt master. Do we want to update it or close it?

PR #186, opened 2022-06-21
Update deps, add tracing

recmo:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/deploy/templates/statefulset.yaml
+++ b/deploy/templates/statefulset.yaml
@@ -66,18 +66,8 @@ spec:
           env:
             - name: LOG_FORMAT
               value: json
-            - name: DD_AGENT_HOST
-              valueFrom:
-                fieldRef:
-                  fieldPath: status.hostIP
-            - name: DD_VERSION
-              valueFrom:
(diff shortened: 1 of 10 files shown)


deploy/templates/statefulset.yaml
```diff
@@ -66,18 +66,14 @@ spec:
...
-            - name: DD_VERSION
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.labels['pod-template-hash']
-            - name: DD_ENV
+            - name: TRACE_OTLP
+              value: "grpc://$(HOST_IP):4317"
+            - name: TRACE_RESOURCE_ENV
               value: {{ .Values.environment }}
-            - name: DD_SERVICE
```
kkowalski, line 79: How is a service name and version determined now ?

recmo: Same way as running it with `--version`. Package name and version.

kkowalski (approved):
LGTM!

PR #197, opened 2022-07-20
Batch insertion sequencing

corydickson:
The goal of this PR is to handle batch insertions of identity commitments by creating several smaller modules that can be scaled. Breaking down the steps, the goal is to divorce the logic of the current endpoint from constructing the merkle tree as well as the interaction with the contract to generate a valid insertion proof.

First, when the insert identity commitment endpoint is called, the hash is inserted into the `pending_commitments` table. This will be used for the next service which will poll this table at some regular interval and attempt to construct a batch of arbitrary size n to be inserted using the new contract interface. 

Once this batch is constructed each identity commitment inside `pending_commitments` will receive a tx hash and a timestamp, latest block number (and account nonce?). If at any point the service fails, it recovers by checking the timestamps/tx/last block in the table and if after some time they are not mined resubmits the batch.  

The client stores the earliest block in the table to begin syncing events from. After the transaction is mined and has at least 20 transactions, commitments are removed from the pending table.


src/database/mod.rs
```diff
@@ -157,4 +157,16 @@ impl Database {
 
         Ok(Hash::default())
     }
+
+    pub async fn pending_commitment(&self, commitment: &Hash) -> Result<usize> {
+        let row = self.pool
+            .execute(sqlx::query(r#"INSERT INTO pending_commitments ( hash ) VALUES ( $1 );"#).bind(uint!(commitment)))
```
recmo, line 163: `uint!` looks like it does nothing and can be removed.

src/database/mod.rs
```diff
@@ -157,4 +157,16 @@ impl Database {
...
+            .await?;
+
+        let idx = row.last_insert_id();
+        match idx {
+            Some(idx) => Ok(idx as usize),
+            None => Err(eyre!("Could not add identity commitment to pending pool"))
+        }
```
recmo, line 170:
```
row.last_insert_id()
   .ok_or_else(|| eyre!("Could not ... pending pool"))
```

https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or_else

recmo (approved):

PR #201, opened 2022-07-26
Add traces

recmo:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/ethereum/mod.rs
+++ b/src/ethereum/mod.rs
@@ -40,7 +40,8 @@ use prometheus::{
 use reqwest::Client as ReqwestClient;
 use std::{error::Error, sync::Arc, time::Duration};
 use thiserror::Error;
-use tracing::{error, info, instrument, warn};
+use tokio::time::timeout;
+use tracing::{error, info, instrument, span, warn, Instrument, Level};
 use url::Url;
 
 const PENDING: Option<BlockId> = Some(BlockId::Number(BlockNumber::Pending));
@@ -113,6 +114,14 @@ pub struct Options {
     /// Polygon mainnet.
     #[clap(long, env, default_value = "31.0")]
     pub min_priority_fee: f64,
+
+    /// Timeout for sending transactions to mempool (seconds).
+    #[clap(long, env, default_value = "30")]
+    pub send_timeout: u64,
+
+    /// Timeout for mining transaction (seconds).
+    #[clap(long, env, default_value = "300")]
+    pub mine_timeout: u64,
 }
 
 // Code out the provider stack in types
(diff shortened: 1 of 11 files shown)


src/ethereum/mod.rs
```diff
@@ -382,21 +406,36 @@ impl Ethereum {
...
         // Send TX to mempool
-        let pending = self
-            .provider
-            .send_transaction(tx.clone(), None)
-            .await
-            .map_err(|error| {
-                error!(?nonce, ?error, "Failed to send transaction");
-                TxError::Send(Box::new(error))
-            })?;
+        let span = span!(Level::INFO, "Send TX to mempool");
```
kkowalski, line 409: Does it create spans with whitespace, if so maybe we could avoid that ?

src/ethereum/mod.rs
```diff
@@ -382,21 +406,36 @@ impl Ethereum {
...
+            error!(?nonce, ?error, "Failed to send transaction");
+            TxError::Send(Box::new(error))
+        })?;
+
         let tx_hash: H256 = *pending;
         info!(?nonce, ?tx_hash, "Transaction in mempool");
+        drop(span_guard);
 
         // Wait for TX to be mined
+        let span = span!(Level::INFO, "Wait for TX to be mined");
```
kkowalski, line 430: I usually like to hide instrumentation as much as possible, here we're introducing multiple span instances within a single function and then we need to follow up with a drop to finish them I presume. Maybe we could introduce some functions to wrap actions  ?

recmo: Yeah, devex could be improved. There is an alternative syntax:

```
info_span!("Wait for TX to be mined").in_scope(|| {
   // ... code in span
})
```

maybe I should use that more. (but also has some downsides, like hard to attach attributes to a span later). I need to experiment with this more.

src/ethereum/mod.rs
recmo: This ends up as the log message (so we need it to be somewhat human readable for the log to make sense). In datadog it looks something like this https://app.datadoghq.com/apm/trace/3176724204366318672?colorBy=service&env=stage&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=15278035570635956810&spanViewType=metadata&timeHint=1658904095107.319

src/ethereum/mod.rs
recmo: In most cases we can just add an `#[instrument]` atrtibute to the function. This one is an exception because there are a lot of complex steps happening that we want to log in detail. I could also split it up in multiple functions.

src/ethereum/mod.rs
recmo: What I don't like is the `opentelemetry.internal` noise everywhere. Maybe we can replace this with the `target` field or something more useful.

kkowalski (approved):

Issue #243, opened 2022-10-14
Cache the merkle tree.


recmo:
On start the sequencer syncs the tree from scratch. This takes a minute and involves many `log_fetch` queries. We should

* Cache the tree in persistent storage.
* Check tree consistency on load.
* Check tree root against chain on start.
* Sync only new leafs.
* Check state root and if invalid either
   * Wipe the state and resync, or
   * Crash.

recmo:
Instead of caching the entire tree we should just cache the log events. Caching closer to raw input reduces the chances of cache corruption. Plus the processing has neglible overhead anyway.

gswirski:
@recmo Caching is implemented now. I'm closing the ticket.

Issue #240, opened 2022-10-14
Split log fetch into smaller block range when the request fails due to timeout.


recmo:
https://docs.alchemy.com/reference/eth-getlogs

The Alchemy `eth_getlogs` call can fail if the request is too large to process. But using too small requests is costly. We have been adjusting the block range size regularly, but depending on Alchemy / Polygon load this arbitrarily fails.

We should dynamically adjust the range. If the request fails with an error indicating it is too large we should retry with a smaller range (say a quarter).

gswirski:
This was implemented in https://github.com/worldcoin/signup-sequencer/pull/283

PR #255, opened 2022-10-31
Events cache

gswirski:
#243 work in progress

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


schemas/database/001_initial.sql
```diff
@@ -8,3 +8,23 @@ CREATE TABLE transactions (
...
+CREATE TABLE logs
+(
+    block_index         BIGINT         NOT NULL,
+    transaction_index   BIGINT         NOT NULL,
+    log_index           BIGINT         NOT NULL,
+    fetch_id            INT            NOT NULL REFERENCES log_fetches (id),
+    group_id            NUMERIC(78, 0) NOT NULL,
+    identity_commitment NUMERIC(78, 0) NOT NULL,
+    root                NUMERIC(78, 0) NOT NULL,
+    UNIQUE (block_index, transaction_index, log_index)
```
recmo, line 29: This is good. Just want to note that using block-number alone is not necessarily sufficient. If a 'reorg' happens the last N blocks get overwritten by a different set of blocks with different transactions/log events. This used to be a lot more common, but with the PoS merge it's about once a week:

https://etherscan.io/blocks_forked

One simple solution that works here is to never sync to the very latest block, but always stay N blocks behind the latest. Where N is the maximum "reorg-depth" that you want to support. Usually `10` is chosen as a conservative maximum.

schemas/database/001_initial.sql
```diff
@@ -8,3 +8,23 @@ CREATE TABLE transactions (
...
+
+CREATE TABLE logs
+(
+    block_index         BIGINT         NOT NULL,
+    transaction_index   BIGINT         NOT NULL,
+    log_index           BIGINT         NOT NULL,
+    fetch_id            INT            NOT NULL REFERENCES log_fetches (id),
+    group_id            NUMERIC(78, 0) NOT NULL,
+    identity_commitment NUMERIC(78, 0) NOT NULL,
+    root                NUMERIC(78, 0) NOT NULL,
```
recmo, line 28: Instead of storing the decoded event we can also store the `raw` event binary data. This decouples the log-caching a bit more from the event type, making it easier to re-use elsewhere.

To make it fully generic you'd need to solve the dependency on the log filter. This can be done through log-topics, but that is getting quite a bit out of scope so let's leave that for when we need it.

schemas/database/001_initial.sql
```diff
@@ -8,3 +8,23 @@ CREATE TABLE transactions (
...
+    first_block BIGINT             NOT NULL,
+    last_block  BIGINT             NOT NULL,
+);
+
+CREATE TABLE logs
+(
+    block_index         BIGINT         NOT NULL,
+    transaction_index   BIGINT         NOT NULL,
+    log_index           BIGINT         NOT NULL,
+    fetch_id            INT            NOT NULL REFERENCES log_fetches (id),
```
recmo, line 25: We don't need this, beacuse 1) I don't see a case for ever wanting to programatically know how we fetched the data. That level of verbosity belongs in logs. 2) We can query `log_fetches` for the block number.

schemas/database/001_initial.sql
```diff
@@ -8,3 +8,23 @@ CREATE TABLE transactions (
     raw         BYTEA                    NOT NULL,
     hash        BYTEA                        NULL
 );
+
+CREATE TABLE log_fetches
```
recmo, line 12: I somehow don't like how this tables turns random whims of the API into a permanent append-only record. One way to make it more palatable is if it could merge adjacent ranges. So when everything is synced up and done this table would just contain one `[0..latest_block]` range spanning the entire chain.

But if all goes well this table will remain small, so no need to add substantial complexity for this.

recmo (commented):
I mentally have a very strong separation between on-chain data (i.e. stuff that is recursively hashed into the latest blockhash) and other data. The event stream is on-chain data and is the ultimate immutable append-only source of truth for the signup-sequencer. On the other hand, the exact sequence and time of `log_fetch`es is basically random noise due to JSON-RPC providers that should be ignored by signup-sequencer as much as possible.

I think this distinction should be clearly preserved in the data structures.

recmo:
> * _**EDIT**: seeing the 48 block reorg now, I'm thinking if we should make it even more robust..._).

That is on Polygon, this will run on Ethereum Mainnet. This parameter depends heavily on the details of the consensus algorithm, which is different for the different chains.

src/app.rs
```diff
@@ -95,14 +88,24 @@ impl App {
     #[allow(clippy::missing_panics_doc)] // TODO
     #[instrument(name = "App::new", level = "debug")]
     pub async fn new(options: Options) -> EyreResult<Self> {
+        if let Ok(app) = Self::bootstrap(options.clone()).await {
+            Ok(app)
+        } else {
+            error!("Error when rebuilding tree from cache. Retrying with db cache busted.");
```
recmo, line 94: This is a catch-all for a very specific error. We shouldn't wipe the database if there was an issue connecting to the Alchemy node.

src/app.rs
```diff
@@ -71,14 +71,7 @@ pub struct Options {
...
 pub struct App {
-    #[cfg(feature = "unstable_db")]
-    #[allow(dead_code)]
-    database: Database,
-
-    #[cfg(not(feature = "unstable_db"))]
-    #[allow(dead_code)]
-    database: (),
-
+    database:    Option<Arc<Database>>,
```
recmo, line 74: No need to `Option` here. With this PR the DB will become a required feature

src/app.rs
```diff
@@ -116,7 +119,12 @@ impl App {
         let merkle_tree = PoseidonTree::new(contracts.tree_depth() + 1, contracts.initial_leaf());
 
         let mut app = Self {
-            database,
+            #[cfg(feature = "unstable_db")]
```
recmo, line 122: No need to feature-gate this, we can make database mandatory.

src/app.rs
```diff
@@ -95,14 +88,24 @@ impl App {
...
+
+            // Remove cached events from database and try again
+            let db = Database::new(options.database.clone()).await?;
+            db.wipe_cache().await?;
+
+            Self::bootstrap(options).await
+        }
+    }
+
+    async fn bootstrap(options: Options) -> EyreResult<Self> {
```
recmo, line 104: Wrapping `new(..)` like this is a hack, we should instead make the wipe-and-retry part of the startup logic.

src/contracts/caching_log_query.rs
```diff
@@ -0,0 +1,214 @@
...
+    cache_blocks_delay: u64,
+    database:           Option<Arc<Database>>,
+}
+
+#[derive(Error, Debug)]
+pub enum Error<ProviderError> {
+    #[error(transparent)]
+    LoadLastBlock(ProviderError),
+    #[error(transparent)]
+    LoadLogs(ProviderError),
```
recmo, line 31: We have two different wrappers for `ProviderError`, but both are `#[transparent]`. This has the consequence that there is no way to distinguish the two cases based on the error message. It would be better to distinguish:

```
    #[error("Error while loading last block: {0}")]
    LoadLastBlock(#[source] ProviderError),
    #[error("Error while loading last block: {0}")]
    LoadLogs(#[source] ProviderError),
```

If we don't use the `#[from]` attribute, we should still set the `#[source]` attribute so the error-chain is preserved.

src/contracts/caching_log_query.rs
```diff
@@ -0,0 +1,214 @@
...
+    #[error("empty log index")]
+    EmptyLogIndex,
+}
+
+impl CachingLogQuery {
+    pub fn new(provider: Arc<ProviderStack>, filter: &Filter) -> Self {
+        Self {
+            provider,
+            filter: filter.clone(),
+            page_size: 10000,
```
recmo, line 49: Is this the same as the `max_log_blocks` option?

It's important that this is configurable, or even better make it dynamic (see #240)

src/contracts/caching_log_query.rs
```diff
@@ -0,0 +1,214 @@
...
+            Some(database) => database.load_logs().await.map_err(Error::Database),
+            None => Ok(vec![]),
+        }
+    }
+
+    async fn fetch_page(&self, filter: Filter) -> Result<Vec<Box<RawValue>>, Error<ProviderError>> {
+        let data: Result<Vec<Box<RawValue>>, ProviderError> = self
+            .provider
+            .provider()
+            .request("eth_getLogs", [&filter])
```
recmo, line 123: Why do it manually instead of `provider.get_logs_paginated` ?

src/contracts/caching_log_query.rs
```diff
@@ -0,0 +1,214 @@
...
+
+            let paginator = Paginator::new(last_block, self.filter.clone(), self.page_size);
+            for page in paginator {
+                let new_events = self.fetch_page(page.filter).await?;
+                for raw_log in new_events {
+                    let log: Log = serde_json::from_str(raw_log.get()).map_err(Error::Parse)?;
+                    if self.should_cache(&log, last_block) {
+                        self.cache_log(raw_log, &log).await?;
+                    }
+                    yield log;
```
recmo, line 89: We should not emit events until they are confirmed.

src/contracts/caching_log_query.rs
```diff
@@ -0,0 +1,214 @@
...
+            .provider
+            .provider()
+            .request("eth_getLogs", [&filter])
+            .await;
+        data.map_err(Error::LoadLogs)
+    }
+
+    fn should_cache(&self, log: &Log, last_block: LastBlock) -> bool {
+        match log.block_number {
+            Some(block) => block + self.cache_blocks_delay <= last_block.eth,
```
recmo, line 130: This delay should apply to the whole app, not just the cache.

src/database/mod.rs
```diff
@@ -1,159 +1,11 @@
...
-use sqlx::{
-    any::AnyKind,
-    migrate::{Migrate, MigrateDatabase, Migrator},
-    pool::PoolOptions,
-    Any, Executor, Pool, Row,
-};
-use tracing::{error, info, instrument, warn};
-use url::Url;
+#[cfg(feature = "unstable_db")]
+mod concrete_database;
```
recmo, line 2: We can make database mandatory. For testing there is the sqlite in-memory database.

src/ethereum/mod.rs
```diff
@@ -105,6 +109,10 @@ pub struct Options {
     #[clap(long, env, default_value = "1000")]
     pub max_log_blocks: usize,
```
recmo, line 110: Is this still being used?

src/ethereum/mod.rs
```diff
@@ -105,6 +109,10 @@ pub struct Options {
...
 
+    /// Minimum number of blocks before events are commited to db cache.
+    #[clap(long, env, default_value = "10")]
+    pub cache_blocks_delay: usize,
```
recmo, line 114: This should apply to the whole module, not just caching. We currently do not do this for transactions, so I opened up #271 for that.

src/ethereum/mod.rs
```diff
@@ -494,18 +504,32 @@ impl Ethereum {
...
             .map_err(Into::into)
+
+        // CachingLogQuery::new(self.provider.provider(), filter)
+        //     .with_page_size(self.max_log_blocks as u64)
+        //     .with_database(database)
+        //     .map_err(Into::into)
+
+        // self.provider
+        //     .get_logs_paginated(filter, self.max_log_blocks as u64)
+        //     .map_err(Into::into)
```
recmo, line 523: Cleanup before merging.

recmo (changes requested):
See comments. Also seeing two clippies:

```
warning: casting to the same type is unnecessary (`u64` -> `u64`)
   --> src/contracts/caching_log_query.rs:102:31
    |
102 |             Some(database) => database.get_block_number().await? as u64,
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `database.get_block_number().await?`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
note: the lint level is defined here
   --> src/lib.rs:2:9
    |
2   | #![warn(clippy::all, clippy::pedantic, clippy::cargo, clippy::nursery)]
    |         ^^^^^^^^^^^
    = note: `#[warn(clippy::unnecessary_cast)]` implied by `#[warn(clippy::all)]`

warning: use Option::map_or instead of an if let/else
   --> src/contracts/caching_log_query.rs:129:9
    |
129 | /         match log.block_number {
130 | |             Some(block) => block + self.cache_blocks_delay <= last_block.eth,
131 | |             None => false,
132 | |         }
    | |_________^ help: try: `log.block_number.map_or(false, |block| block + self.cache_blocks_delay <= last_block.eth)`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
note: the lint level is defined here
   --> src/lib.rs:2:55
    |
2   | #![warn(clippy::all, clippy::pedantic, clippy::cargo, clippy::nursery)]
    |                                                       ^^^^^^^^^^^^^^^
    = note: `#[warn(clippy::option_if_let_else)]` implied by `#[warn(clippy::nursery)]`
```

schemas/database/002_event_log.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE logs
+(
+    block_index         BYTEA NOT NULL,
+    transaction_index   BYTEA NOT NULL,
+    log_index           BYTEA NOT NULL,
```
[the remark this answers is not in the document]
recmo: For block_number to exceed 2^64 within the next 100 years, a blockchain will need to produce blocks at over 5GHz. We won't see that kind of performance anytime soon.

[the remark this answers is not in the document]
gswirski: My problem with `INT` and `BIGINT` is that they don't match the definition of `Log` in `ethers_core`:
```
    pub block_number: Option<U64>,
    pub transaction_index: Option<U64>,
    pub log_index: Option<U256>,
```

In postgres, `INT` is i32 (vs u64 in `Log`) and `BIGINT` is i64 (vs u256 in `Log`).

I agree that `BYTEA` might be a little wasteful but since we are keeping raw logs anyway, it seemed more fitting?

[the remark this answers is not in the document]
recmo: I'm worried that using a non-integer type will make it hard to do range-based queries which we might want to do in the future for various reasons.

Note that Ethereum itself and the underlying JSON-RPC doesn't actually specify the range, just a "quantity" to be encoded in hex. It is in principle unbounded. https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getlogs

The types in `ethers-core` are quite arbitrary. For comparison, the (now deprecated) open-ethereum (fka parity) client uses `usize`: https://github.com/openethereum/openethereum/blob/6c2d392d867b058ff867c4373e40850ca3f96969/crates/ethcore/types/src/log_entry.rs#L62

So really everyone just picks a reasonably sized type here (the choice of `U256` in ethers-core actually feels wrong to me. U256 is the i-dont-want-to-think-about-it-default in Ethereum ecosystem, so maybe just laziness).

I gave an argument above for the `block_number`, the  other two `*_index` fields are indices in arrays that are bounded by the block size and gas limit, so very small as well.

An argument could be made that we want to be able to cache all the values that can be expressed in `ethers-core` types. I don't think that is very relevant, given that `ethers-core` itself can not represent all the values that could in theory be returned from the `eth_getLogs` call (which are unbounded integers).

src/contracts/caching_log_query.rs
gswirski: Yes, `page_size` is set to `max_log_blocks` in https://github.com/worldcoin/signup-sequencer/pull/255/files#diff-6f2d98d6ba19d324003887e0fbb816a48b66ac232e041924824039ac5116a729R510

10000 is the default value from `get_logs_paginated`.

src/ethereum/mod.rs
gswirski: https://github.com/worldcoin/signup-sequencer/pull/255#discussion_r1020286663 - it's still used to configure `page_size`

src/ethereum/mod.rs
gswirski: I've renamed this option to `confirmation_blocks_delay` should be a bit more fitting name for #271. However, I think #271 should be out of scope of this PR. It will depend heavily on @kustosz work on #258.

src/contracts/caching_log_query.rs
gswirski: Reverted it back to `provider.get_logs_paginated` but I'm not entirely sure which way is better.

Using `provider.get_logs_paginated` saves us some code (no more `Paginator` struct) and keeps things on higher abstraction layer (`Log` vs `Box<RawValue>`).

On the other hand, now we have to serialize `Log`s manually when storing them in db (see: https://github.com/worldcoin/signup-sequencer/pull/255/files#diff-abf9484cd24d8332d6292b4efa6f936372717653df66b551131ba70c9272d780R95). With us doing requests directly, we would store the original raw json log and deserialize once for stream consumers. Now, we get deserialized objects from `get_logs_paginated` and then serialize it again to store in db.

src/contracts/app.rs
```diff
@@ -0,0 +1,480 @@
...
+                ?error,
+                "Root mismatch between tree and contract."
+            );
+            return Err(ServerError::RootMismatch);
+        }
+
+        Ok(InclusionProofResponse { root, proof })
+    }
+
+    /// Stores the Merkle tree to the storage file.
```
gswirski, line 266: @recmo this comment seems incorrect?

recmo: It is incorrect. I think this is left over from a previous function that actually did store the tree.

src/ethereum/mod.rs
recmo: Agreed, it is out of scope for this PR.

recmo (approved):

Issue #258, opened 2022-10-31
Work queue for `insert_identity`


recmo:
Right now the `insert_identity` API call is fully synchronous:

1. It receives an HTTP POST on `/insertIdentity`.
2. It parses and validates the request.
3. It takes the global tree lock (exclusive write).
4. It checks for duplicate insertions.
5. It creates and signs an Ethereum transaction.
6. It waits for the transaction to be mined.
7. It updates the internal tree representation.
8. It downgrades the lock to a non-exclusive read lock.
9. It verifies the tree-root against the on-chain one.
10. It releases the global tree lock.
11. It returns a response to the API request.

The exclusive global lock prevents concurrent request handling and steps 5-6 can take a lot of time and/or fail. 

---

Instead we want to get rid of the lock, create a work-queue and split this up in three separate threads, A, B, C:

A:
1. A receives an HTTP POST on `/insertIdentity`.
2. A parses and validates the request.
4. A checks for duplicate insertions. (TBD: How?)
5. A inserts the identity in the work-queue.
6. A returns a response to the request.

B:
1. It takes an item from the work-queue.
2. It flags it as in-progress.
2. It creates and signs an Ethereum transaction.
3. It waits for the transaction to be mined.
4. Marks the work item as done.

C:
1. Monitors the events.
2. Updates the internal tree on new events.
3. Checks the tree-root for consistency.
4. TBS: Check the events against the work-queue for consistency?

gswirski:
Done in #284

PR #264, opened 2022-11-07
Pending identity inserts

kustosz:
## Motivation
Part of the re-arch project, this code corresponds to processes a. and b. as defined in #258 . Process C is to be cut out based on #270.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/app.rs
```diff
@@ -476,6 +479,144 @@ impl App {
...
+
+        Ok(())
+    }
+
+    #[instrument(level = "debug", skip_all)]
+    fn spawn_identity_committer(self: Arc<Self>) {
+        let mut listener = self.event_bus.subscribe();
+        tokio::spawn(async move {
+            let mut transaction_in_progress = false;
+            let (tx, mut rx) = tokio::sync::mpsc::channel(1);
```
recmo, line 575: These names are not descriptive. I'm also not sure if `mspc` is the appropriate channel type.

src/app.rs
```diff
@@ -476,6 +479,144 @@ impl App {
...
+            let mut work_queue: VecDeque<(usize, Hash)> = VecDeque::new();
```
recmo, line 576: Isn't this redundant with the queue in the `event_bus` channel and the new database table?

src/app.rs
```diff
@@ -476,6 +479,144 @@ impl App {
...
+                        for msg in msgs {
+                            let Event::PendingIdentityInserted { group_id, commitment } = msg;
+                            info!("Identity committer received event group_id: {}, identity: {}.", group_id, commitment);
+                            work_queue.push_back((group_id, commitment));
+                        }
+                    },
+                    else => {}
+                }
+            }
+        });
```
recmo, line 618: Why not

```
while let Some(event) = listener.recv().await {
 app.commit_identity(...
}
```
?

recmo: Or even simpler, just use the database as work queue. And poll the database in a loop.

To make it reactive we can have it stop when there is no more work and use an event to start the loop. This can be a simple [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html) and doesn't have to carry state like the event_bus.

recmo (changes requested):

recmo (changes requested):
I'm a bit concerned with the work queue being stored in three different places.

1. In the database.
2. In the event bus channel.
3. In the work_queue VecDeque

I think we should stick to using the database exclusively. That should simplify things and avoid potential issues with keeping state in sync.

src/timed_read_progress_lock.rs
```diff
@@ -0,0 +1,183 @@
...
+
+/// A 3-stage lock, with the following stages:
+/// 1. Read – can be held by multiple users at the same time.
+/// 2. Progress – can be held by a single user, but does not disallow reads.
+/// 3. Write – can be held by a single user, that must have acquired the
+/// progress lock before, precludes any read locks.
+///
+/// Can be used to specify a critical modification section, where certain
+/// computations should be exclusive, even if the structure can still be read
+/// (possibly getting stale results), narrowing the read-excluding section.
```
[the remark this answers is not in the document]
kustosz: I don't think so, at least I'm not aware of this. It's just a primitive I needed but couldn't find :)

src/app.rs
```diff
@@ -476,6 +479,144 @@ impl App {
...
+            panic!("Sequencer potentially deadlocked, terminating.");
+            #[allow(unreachable_code)]
+            e
+        })?;
+
+        // Update  merkle tree
+        tree.set(identity_index, commitment);
+
+        // Downgrade write lock to progress lock
+        let tree = tree.downgrade_to_progress();
```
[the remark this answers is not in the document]
kustosz: This looks like a serious sanity check, so I don't think it's a good idea to allow something to run concurrently from a broken state. Or if we're OK with another job picking up a broken state, maybe we should remove this check altogether (and allow the next insert to fail).

kustosz:
@recmo you're right, it was better to just poll the DB. Also I ended up holding on to the thread handle, so it can be supervised / restarted in the future. I think I addressed all of your feedback, either by changing code or following up in the comments.

src/database.rs
```diff
@@ -125,6 +125,82 @@ impl Database {
...
+                sqlx::query(
+                    r#"INSERT INTO pending_identities (group_id, commitment)
+                           VALUES ($1, $2);"#,
+                )
+                .bind(group_id as i64)
+                .bind(identity),
+            )
+            .await?;
+        Ok(())
+    }
```
recmo, line 144: Stylistic nitpick: I think it looks better with an intermediate variable:
```rust
        let query = sqlx::query(
            r#"UPDATE pending_identities
                    SET mined_in_block = $1, assigned_leaf_idx = $2
                    WHERE group_id = $3 AND commitment = $4;"#,
        )
        .bind(block_number as i64)
        .bind(tree_idx as i64)
        .bind(group_id as i64)
        .bind(commitment);
        Ok(())
```
This reduces the amount of identation and lines just containing `)`.

src/database.rs
```diff
@@ -125,6 +125,82 @@ impl Database {
...
+            .pool
+            .fetch_optional(sqlx::query(
+                r#"SELECT group_id, commitment
+                           FROM pending_identities
+                           WHERE mined_in_block IS NULL
+                           ORDER BY created_at ASC
+                           LIMIT 1;"#,
+            ))
+            .await?;
+        Ok(row.map(|row| (usize::try_from(row.get::<i64, _>(0)).unwrap(), row.get(1))))
```
recmo, line 201: It's a bit annoying that there is no `usize` impl, but understandable since some databases (i.e. postgres) don't have a `u64` type.

FWIW, this is a bit more concise `row.get::<i64, _>(0).try_into().unwrap()`

src/identity_committer.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+
+    async fn shutdown(self) -> eyre::Result<()> {
+        info!("Sending a shutdown signal to the committer.");
+        // Ignoring errors here, since we have two options: either the channel is full,
+        // which is impossible, since this is the only use, and this method takes
+        // ownership, or the channel is closed, which means the committer thread is
+        // already dead.
+        self.shutdown_sender.send(()).await.ok();
```
recmo, line 46: `.ok()` just returns an `Option<T>` containing the value, if any. If we want to dispose of the result without triggering a panic or the `must_use` lint the pattern is:

```rust
let _ = self.shutdown_sender.send(()).await;
```

src/identity_committer.rs
```diff
@@ -0,0 +1,204 @@
...
+            contracts,
+            tree_state,
+        }
+    }
+
+    #[instrument(level = "debug", skip_all)]
+    pub async fn start(&self) {
+        let mut instance = self.instance.write().await;
+        if instance.is_some() {
+            info!("Identity committer already running");
```
recmo, line 77: This should at least be a warning, and it is arguably an error if this happens.

src/identity_committer.rs
```diff
@@ -0,0 +1,204 @@
...
+            return;
+        }
+        let (shutdown_sender, mut shutdown_receiver) = mpsc::channel(1);
+        let (wake_up_sender, mut wake_up_receiver) = mpsc::channel(1);
+        let database = self.database.clone();
+        let tree_state = self.tree_state.clone();
+        let contracts = self.contracts.clone();
+        let handle = tokio::spawn(async move {
```
recmo, line 85: This basically creates a closure taking the variables cloned above. In the future when this gets more complex we probably want a explicit struct for the context like so:

```rust
struct Task {
  database: // ...
  // ..
}

impl Task {
   async fn run(mut self) { ... }
}
```

And then launch it by 

```
let task = Task { ... };
let handler = tokio::spawn(task.run());
```

src/identity_committer.rs
```diff
@@ -0,0 +1,204 @@
...
+                error!(
+                    computed_root = ?tree.root(),
+                    ?error,
+                    "Root mismatch between tree and contract."
+                );
+                error
+            })?;
+
+        // Immediately write the tree to storage, before anyone else can
+        // TODO: Store tree in database
```
recmo, line 178: I think this comment is copied over, but no longer relevant? At any rate the TODO is captured in one of the open issues, so we can remove it here.

src/identity_committer.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        let block = receipt
+            .block_number
+            .expect("Transaction is mined, block number must be present.");
+        info!(
+            "Identity inserted in block {} at index {}.",
+            block, identity_index
+        );
+        database
+            .mark_identity_inserted(group_id, &commitment, block.as_usize(), identity_index)
```
recmo, line 161: This works as long as there is a single `identity_comitter`.
If there is more than one we need to transactionally fetch a batch and flag them as 'in progress'. That is out of scope for this PR.
But we should at least add a warning in the `IdentityCommitter`  documentation (e.g. a `///` comment).

recmo (approved):

src/app.rs
```diff
@@ -71,14 +72,32 @@ pub struct Options {
     pub lock_timeout: u64,
 }
 
+pub struct TreeState {
+    pub next_leaf:   AtomicUsize,
+    pub merkle_tree: TimedReadProgressLock<PoseidonTree>,
+}
```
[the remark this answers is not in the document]
kustosz: Good point, strange how it was allowed to be changed independently.

src/identity_committer.rs
kustosz: I'll bump it to a warning, I don't think it is an error. In the future, we may want to restart the committer and this behavior allows the first call that spots it's dead to do so safely – if someone else already restarted, it'll just be a no-op.

Issue #242, opened 2022-10-14
Move production config to /deploy/values.yaml


recmo:
We should avoid having config in the .github folder and move it to /deploy

PR #265, opened 2022-11-07
Move production config to /deploy/values.yaml (#242)

dcbuild3r:
## Motivation

Issue #242

## Solution

creating separate files in `deploy/` to set env variables for different helm deploys.

## PR Checklist

- [x] Test locally and see if it works
- [x] Verify this is the helm way to do things


.github/workflows/build-test-deploy.yml
```diff
@@ -395,14 +373,5 @@ jobs:
         run: helm upgrade "${{ github.event.repository.name }}" ./deploy
           --install --timeout 60s
           --namespace "${{ github.event.repository.name }}"
-          --set image.tag="${{ github.sha }}"
```
kszarek, line 398: I'd recommend leaving it is. There is no `github.sha` variable defined in the Helm Chart.

dcbuild3r: Ok, the other variables are also defined in the github actions context no? Things like secrets etc.

recmo (approved):

PR #283, opened 2022-11-17
dynamic page size for fetching events

gswirski:
Implements https://github.com/worldcoin/signup-sequencer/issues/240

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

PR #284, opened 2022-11-18
Integrate event subscriber with identity committer queue

gswirski:
This PR is dependent on https://github.com/worldcoin/signup-sequencer/pull/264 & https://github.com/worldcoin/signup-sequencer/pull/283. It also supersedes https://github.com/worldcoin/signup-sequencer/pull/270.

## To Do
- [x] reconcile pending identities work queue
- [x] add tests
- [x] remove progress lock?

## Motivation

Implements the final piece of https://github.com/worldcoin/signup-sequencer/issues/258


gswirski:
@kustosz this is now ready for review. While you are looking at it, I'll continue searching for a better solution to those random `sleep`s in tests.

It seems like we no longer need your `progress` state of the `TimedReadProgressLock`?

src/database.rs
```diff
@@ -145,21 +145,50 @@ impl Database {
...
-        .bind(tree_idx as i64)
+        .bind(relative_index)
         .bind(group_id as i64)
         .bind(commitment);
+
         self.pool.execute(query).await?;
         Ok(())
     }
 
+    pub async fn pending_identity_confirmed(
```
kustosz, line 164: 2 things wrong with this IMO:

1. It lacks an active verb in the name – something like `mark_pending_identity_confirmed` or something to that effect – this name doesn't tell me "I'm about to mutate something".
2. It hides a lot of complexity behind this name – the current name only allows you to remove a pending item, not to invalidate all previous names. Change the name of this, or split it into two methods, so that I'm not surprised when an innocuously named method queues a bunch of work.

src/ethereum/mod.rs
```diff
@@ -522,6 +526,15 @@ impl Ethereum {
         Ok(receipt)
     }
 
+    pub async fn confirmed_block_number(&self) -> Result<U64, EventError> {
+        self.provider
+            .provider()
+            .get_block_number()
+            .await
+            .map(|num| num.saturating_sub(U64([self.confirmation_blocks_delay as u64])))
```
kustosz, line 534: there's a `From<u64>` for `U64`

src/ethereum_subscriber.rs
```diff
@@ -0,0 +1,388 @@
...
+        info!(
+            starting_block,
+            end_block, "processing events in ethereum subscriber"
+        );
+
+        let mut events = contracts
+            .fetch_events(starting_block, Some(end_block), database.clone())
+            .boxed();
+
+        let shutdown = await_shutdown();
```
kustosz, line 130: I prefer to use a chan for this and pipe from `await_shutdown` explicitly – gives you more control in weird setups, like testing.

src/identity_committer.rs
```diff
@@ -128,61 +120,33 @@ impl IdentityCommitter {
...
-        );
+
+        info!("Identity submitted in block {}.", block);
         database
-            .mark_identity_inserted(group_id, &commitment, block.as_usize(), identity_index)
+            .mark_identity_inserted(
+                group_id,
+                &commitment,
+                block.as_usize(),
+                index.fetch_add(1, Ordering::AcqRel),
```
kustosz, line 143: This is mighty strange, a counter that is not at all restart-resilient? Why do we care at all?

kustosz: What I'm saying is: remove it from the DB, what uses this?

kustosz (approved):
Good stuff in general, some nitpicks left inline

src/database.rs
gswirski: Renamed to `confirm_identity_and_retrigger_stale_recods`

src/identity_committer.rs
gswirski: You are right, this was completely wrong. I removed the index column and simply depend on created_at order instead. This should be simpler and restart-resilient.

src/ethereum_subscriber.rs
gswirski: Removed. Ethereum subscriber shouldn't care about system interrupts.

kustosz (approved):

PR #301, opened 2022-11-25
Don't lock the tree accross network operations

kustosz:
## Motivation
Hiccups in service availability may render the internal state impossible to modify and cause crashes. Let's not lock the tree accross them.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -201,10 +201,8 @@ impl App {
             return Err(ServerError::InvalidGroupId);
         }
 
-        let tree = self.tree_state.read().await?;
-
         if commitment == self.contracts.initial_leaf() {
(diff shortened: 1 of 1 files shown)


gswirski (approved):

src/app.rs
```diff
@@ -217,19 +215,22 @@ impl App {
             .pending_identity_exists(group_id, &commitment)
             .await?
```
recmo, line 216: Doesn't this have a race condition where two concurrent identical insert requests could both pass this point?

Out of scope for this PR, but let's open an issue for it.

recmo (approved):

PR #304, opened 2022-11-26
Catch panics and errors in background threats

recmo:
If the future supplied to `tokio::spawn` returns an error or panics this is not noticed until the task handle is joined. This PR makes it such that the fault is loggedn and the whole process aborts immediately.


recmo:
Build and passes tests on stable, but I still get ICE on latest nightly

recmo:
Basically hitting this issue: https://github.com/primait/prima_tracing.rs/pull/49

gswirski (approved):

PR #307, opened 2022-11-29
replace eyre with anyhow

gswirski:
Following up on https://github.com/worldcoin/signup-sequencer/pull/295, we had another instance where logging `eyre::Report` with `tracing` and `otlp` enabled completely hanged the app (and prevented it from restarting successfully).

It looks like `anyhow` doesn't have this problem. We can continue investigating what exactly causes OTLP and eyre to misbehave so badly, but for now this should fix the production.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -3,9 +3,9 @@ use crate::{
     database,
 };
 use ::prometheus::{opts, register_counter, register_histogram, Counter, Histogram};
+use anyhow::{bail, ensure, Context, Error as EyreError, Result as AnyhowResult};
 use clap::Parser;
-use cli_batteries::{await_shutdown};
-use anyhow::{anyhow, bail, ensure, Error as EyreError, Result as AnyhowResult};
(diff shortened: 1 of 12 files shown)


src/server.rs
```diff
@@ -173,7 +173,7 @@ where
 
 #[instrument(level="info", name="api_request", skip(app), fields(http.uri=%request.uri(), http.method=%request.method()))]
 async fn route(request: Request<Body>, app: Arc<App>) -> Result<Response<Body>, hyper::Error> {
-    trace_from_headers(request.headers());
+    // trace_from_headers(request.headers());
```
recmo, line 176: This is orthogonal to eyre, but makes sense to remove since it also touches OTLP/tracing and could cause trouble.

recmo (approved):
Good as quick fix, but we should root cause this and probably get eyre with span traces to work correctly.

PR #364, opened 2023-01-26
Initial Refactoring for Batching Integration

iamrecursion:
This PR refactors how the existing legacy contract is handled by introducing a trait to unify behaviour between the two contracts. This is in preparation for introducing the batching contract to the codebase with the swap between the two controlled by a cargo feature.

There are no functional changes in this PR, but it does make some renames around the codebase to make things clearer now that "semaphore" is not the only contract. To that end, it intentionally does not test what little integration of the batching prover there is.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,4 @@
 [package]
-# Please adjust all these values, including the license
 name = "signup-sequencer"
 version = "0.1.0"
 authors = [
@@ -9,14 +8,15 @@ edition = "2021"
 build = "build.rs"
 homepage = "https://github.com/worldcoin/signup-sequencer"
 repository = "https://github.com/worldcoin/signup-sequencer"
-description = "Template for a Rust CLI tool"
+description = "A tool that processes WorldID signups on-chain."
 keywords = ["worldcoin", "protocol", "signup"]
 categories = ["cryptography::cryptocurrencies"]
 readme = "Readme.md"
 license-file = "mit-license.md"
 
 [features]
-default = [ ]
+default = []
+batching-contract = []
 bench = ["criterion", "proptest"]
 mimalloc = ["cli-batteries/mimalloc"]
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -27,29 +27,29 @@ path = "criterion.rs"
 required-features = ["bench", "proptest"]
 
 [dependencies]
-anyhow = "1.0.68"
+anyhow = { version = "1.0.68", features = ["backtrace"] }
```
gswirski, line 30: If `backtrace` feature is not required for this PR (I doubt it is), let's remove it from here. I'd like to test this particular change separately.

I know it can be useful, but we had some terrible time with eyre + tracing + opentelemetry (the sequencer would just hang on the `info!` macro).

src/contracts/batching/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        Self: Sized,
+    {
+        // Check that there is code deployed at the target address.
+        let address = options.identity_manager_address;
+        let code = ethereum.provider().get_code(address, None).await?;
+        if code.as_ref().is_empty() {
+            error!(
+                ?address,
+                "No contract code is deployed at the provided address."
+            )
```
gswirski, line 45: Clippy wants a `;`

src/contracts/batching/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+            abi,
+        };
+
+        Ok(identity_manager)
+    }
+
+    fn tree_depth(&self) -> usize {
+        // TODO [Ara] Work out what to do for this. There is no sane answer for a
+        //   contract that can work with arbitrary verifiers.
+        todo!()
```
gswirski, line 79: This could be a regular config option provided on startup via `clap`. No action for now.

src/contracts/mod.rs
```diff
@@ -1,228 +1,112 @@
...
+    /// contract on the chain.
+    async fn assert_latest_root(&self, root: Field) -> anyhow::Result<()>;
+
+    /// Asserts that the provided `root` is a valid root.
+    ///
+    /// A valid root is one that has not expired based on the time since it was
+    /// inserted into the history of roots on chain.
+    async fn assert_valid_root(&self, root: Field) -> anyhow::Result<()>;
+
+    // TODO [Ara] Consider refactoring this to remove the stream when possible.
```
gswirski, line 98: No need for refactor. Once OZ Relay is merged https://github.com/worldcoin/signup-sequencer/pull/357, we can replace `fetch_events` with some other way of confirming transactions.

src/contracts/mod.rs
```diff
@@ -1,228 +1,112 @@
...
-        let error = result.to_string();
-        if error.contains("0x09bde339") {
-            return Ok(());
-        }
-        if error.contains("0x504570e3") {
-            return Err(anyhow!("Invalid root"));
-        }
-        Err(anyhow!("Error verifiying root: {}", result))
-    }
+    ) -> Option<Pin<Box<dyn Stream<Item = Result<Log<MemberAddedEvent>, EventError>> + Send + '_>>>;
```
gswirski, line 108: Clippy is unhappy and wants some type aliases.

src/contracts/mod.rs
```diff
@@ -1,228 +1,112 @@
...
 }
+
+/// A type for an identity manager object that can be sent across threads.
+pub type ThreadableIdentityManager = dyn IdentityManager + Send + Sync;
```
gswirski, line 112: This name makes sense, but trips me a little since we never explicitly create threads.

Maybe change to `SharedIdentityManager = Arc<dyn IdentityManager + Send + Sync>` to match `SharedTreeState`? That one reference in the `IdentityCommitter` could be replaced with `Arc`.

gswirski (approved):
Looks great. Minor comments, the biggest one being the addition of "backtrace" feature in "anyhow" crate.

Cargo.toml
iamrecursion: Ah, good catch! I hadn't intended to leave this on. I enabled it for debugging. :D

PR #377, opened 2023-02-01
Pending merkle proofs

kustosz:
This allows the app to return a merkle inclusion proof right after insertion. It changes the real source of truth for the app – now the "master copy" is the database. This will play nice with the new contract, where we have full control over leaf ordering, even with ethereum's instability as a database. A lot of code is now removed – some of it will never come back, some of it will come back in a new form when the new contract is integrated.

Keep in mind that the way we currently represent the tree is very rudimentary and memory hungry. It will be changed in the near future, when we want to increase tree depth. We'll have to rewrite the current implementation of the tree to be more efficient then, but it should not be a blocker at current tree depths.


src/app.rs
```diff
@@ -114,100 +91,49 @@ impl App {
...
-        let tree_state = Arc::new(TimedRwLock::new(
-            Duration::from_secs(options.lock_timeout),
-            TreeState::new(
-                identity_manager.tree_depth() + 1,
-                identity_manager.initial_leaf_value(),
-            ),
-        ));
+        let tree_state = Self::initialize_tree(
+            &database,
+            identity_manager.tree_depth() + 1,
```
iamrecursion, line 96: Probably worth retaining a comment as to why this is a magic number.

src/app.rs
```diff
@@ -114,100 +91,49 @@ impl App {
...
             database.clone(),
             identity_manager.clone(),
             tree_state.clone(),
         ));
-        let chain_subscriber = EthereumSubscriber::new(
-            options.starting_block,
-            database.clone(),
-            identity_manager.clone(),
-            tree_state.clone(),
-        );
```
iamrecursion, line 136: With the remover of this, it'd be great if you can remove the `fetch_events` method from the `IdentityManager` trait. That's been a pain and not having to deal with it would be wonderful.

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+    pub status: Status,
+    pub root:   Field,
+    pub proof:  Proof,
+}
+
+#[derive(Clone)]
+pub struct TreeState {
+    mined:  TreeVersion,
+    latest: TreeVersion,
+}
```
iamrecursion, line 198: I would add a todo comment for the additional version explaining why it doesn't exist yet.

src/database/mod.rs
```diff
@@ -130,229 +126,103 @@ impl Database {
         Ok(Self { pool })
     }
 
-    pub async fn insert_pending_identity(
+    pub async fn insert_identity_if_not_duplicate(
```
iamrecursion, line 129: I'm not sure this name captures that it's a nop on a duplicate identity. Perhaps `insert_identity_or_nop_if_duplicate`?

src/database/mod.rs
```diff
@@ -130,229 +126,103 @@ impl Database {
...
-        .bind(IDENTITY_SUBMITTING); // new status
-
+        .bind(leaf_index as i64)
+        .bind(<&str>::from(Status::Mined));
         self.pool.execute(query).await?;
         Ok(())
     }
 
-    pub async fn mark_identity_inserted(
+    pub async fn get_updates_range(
```
iamrecursion, line 160: Perhaps `get_updates_in_range`? Seems a touch clearer to me.

src/database/mod.rs
```diff
@@ -130,229 +126,103 @@ impl Database {
...
+            })
+            .collect::<Result<Vec<_>, _>>()
     }
 
-    pub async fn delete_pending_identity(
-        &self,
-        group_id: usize,
-        commitment: &Hash,
-    ) -> Result<(), Error> {
+    pub async fn get_identity_index(&self, identity: &Hash) -> Result<Option<TreeItem>, Error> {
```
iamrecursion, line 186: This name is also a bit strange. Am I right in understanding that it gets the index in the tree of the provided `identity`? If that's the case, perhaps `get_identity_index_in_tree`?

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+
+    #[must_use]
+    pub fn get_latest_tree(&self) -> TreeVersion {
+        self.latest.clone()
+    }
+
+    #[must_use]
+    pub fn get_mined_tree(&self) -> TreeVersion {
+        self.mined.clone()
+    }
```
iamrecursion, line 214: I'd rename these to something that indicates that it's a clone under the hood. It confused me at the start until I looked at the impl.

src/identity_committer.rs
```diff
@@ -94,12 +96,11 @@ impl IdentityCommitter {
         let (wake_up_sender, mut wake_up_receiver) = mpsc::channel(1);
         let database = self.database.clone();
         let identity_manager = self.identity_manager.clone();
-        let tree_state = self.tree_state.clone();
+        let tree = self.tree_state.get_mined_tree();
```
iamrecursion, line 99: This needs a clearer name. `mined_tree`?

src/identity_committer.rs
```diff
@@ -118,15 +119,12 @@ impl IdentityCommitter {
     async fn process_identities(
         database: &Database,
         identity_manager: &(dyn IdentityManager + Send + Sync),
```
iamrecursion, line 121: Probably worth pulling this out into a type synonym in `contracts/mod.rs`. `DynIdentityManager`. Then you can take a `&DynIdentityManager`.

src/identity_committer.rs
```diff
@@ -118,15 +119,12 @@ impl IdentityCommitter {
...
-        tree_state: &SharedTreeState,
+        tree: &TreeVersion,
```
iamrecursion, line 122: I wanna see this name be more specific about the kind of tree it is: `mined_tree`.

src/identity_committer.rs
```diff
@@ -138,49 +136,23 @@ impl IdentityCommitter {
     async fn commit_identity(
         database: &Database,
         identity_manager: &(dyn IdentityManager + Send + Sync),
-        tree_state: &SharedTreeState,
-        group_id: usize,
-        commitment: Hash,
+        tree: &TreeVersion,
```
iamrecursion, line 139: Same here.

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
 
 pub type Hash = <PoseidonHash as Hasher>::Hash;
 
-pub struct TreeState {
-    pub next_leaf:   usize,
-    pub merkle_tree: PoseidonTree,
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub struct TreeUpdate {
+    pub leaf_index: usize,
+    pub element:    Hash,
```
iamrecursion, line 17: If these are public they should have documentation.

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+        });
+    }
+
+    fn update_without_diff(&mut self, leaf_index: usize, element: Hash) {
+        self.tree.set(leaf_index, element);
+        self.next_leaf = leaf_index + 1;
+    }
+}
+
+#[derive(Clone)]
```
iamrecursion, line 92: Derive `Debug` at least please!

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+    #[must_use]
+    pub const fn new(leaf_index: usize, element: Hash) -> Self {
+        Self {
+            leaf_index,
+            element,
+        }
+    }
+}
+
+struct TreeVersionData {
```
iamrecursion, line 30: At least derive `Debug`.

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+                .proof(leaf)
+                .expect("impossible, tree depth mismatch between database and runtime"),
+        )
+    }
+}
+
+pub struct TreeItem {
+    pub status:     Status,
+    pub leaf_index: usize,
+}
```
iamrecursion, line 152: Derive `Debug` please.

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+
+#[derive(Clone, Copy, Serialize)]
```
iamrecursion, line 154: `Debug`

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+impl From<Status> for &str {
+    fn from(scope: Status) -> Self {
+        match scope {
+            Status::Pending => "pending",
+            Status::Mined => "mined",
+        }
+    }
+}
+
+#[derive(Serialize)]
```
iamrecursion, line 186: `Debug`.

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+    pub fn get_latest_tree(&self) -> TreeVersion {
+        self.latest.clone()
+    }
+
+    #[must_use]
+    pub fn get_mined_tree(&self) -> TreeVersion {
+        self.mined.clone()
+    }
+
+    pub async fn get_proof(&self, item: &TreeItem) -> InclusionProof {
```
iamrecursion, line 216:
```suggestion
    pub async fn get_proof_for(&self, item: &TreeItem) -> InclusionProof {
```

src/identity_tree.rs
```diff
@@ -1,26 +1,247 @@
...
+        let (root, proof) = tree.get_proof(item.leaf_index).await;
+        InclusionProof {
+            status: item.status,
+            root,
+            proof,
         }
     }
 }
+
+pub struct CanonicalTreeBuilder(TreeVersionData);
```
iamrecursion, line 230: Derive `Debug`.

iamrecursion (approved):
Mostly cosmetic and naming things. Looks good otherwise!

src/identity_tree.rs
kustosz: I thought about it, but in the end I've decided the whole thing should be pretty number-of-versions-agnostic. I can see it being anywhere between 2 and 5 in the coming months.

src/database/mod.rs
kustosz: Changed to `insert_indentity_if_does_not_exist`. `nop` in a name sounds really bad to me.

src/database/mod.rs
kustosz: changed to `get_identity_leaf_index`, leaf index is what we consistently use for this

src/identity_tree.rs
kustosz: It's a clone of an `Arc`, why would you care? I'm assuming if someone is reading these implementations, they need to understand the internal types. From the outside, you just get a reference to the tree.

src/identity_committer.rs
kustosz: I'm assuming this will be monomorphised very soon though?

src/identity_tree.rs
kustosz: I have trouble understanding what needs documenting here.

src/identity_tree.rs
kustosz: Can't, `PoseidonTree` is an upstream dep and it doesn't do `Debug`. Will change in the future, when I touch that codebase.

src/identity_tree.rs
iamrecursion: Okay, sounds good!

src/identity_tree.rs
iamrecursion: Ah, I missed that it was an `Arc`! All good.

src/identity_tree.rs
iamrecursion: Noted!

src/identity_tree.rs
iamrecursion: It's not clear to me what `leaf_index` actually entails. Is it the index at which the `element` is inserted?

src/identity_committer.rs
iamrecursion: That's true. Leave it.

src/identity_tree.rs
kustosz: Yeah. A merkle tree holds elements in leaves only. And they are indexed. It's consistent with the usage in the database and in the contract too.

PR #378, opened 2023-02-02
Connect to and work with the prover service

iamrecursion:
This commit does some initial integration with the batch prover by connecting to `semaphore-mtb` and requesting proofs from it. It does not yet use these proofs in the batch prover.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,11 +1,15 @@
 use crate::{
     contracts,
-    contracts::{legacy::Contract as LegacyContract, IdentityManager, SharedIdentityManager},
+    contracts::{
+        batching::Contract as BatchingContract, legacy::Contract as LegacyContract,
+        IdentityManager, SharedIdentityManager,
+    },
     database::{self, Database},
     ethereum::{self, Ethereum},
     ethereum_subscriber::{Error as SubscriberError, EthereumSubscriber},
     identity_committer::IdentityCommitter,
     identity_tree::{Hash, SharedTreeState, TreeState},
+    prover,
     server::{Error as ServerError, ToResponseCode},
     timed_rw_lock::TimedRwLock,
 };
@@ -64,6 +68,9 @@ pub struct Options {
     #[clap(flatten)]
     pub database: database::Options,
 
+    #[clap(flatten)]
+    pub prover: prover::Options,
+
     /// Block number to start syncing from
     #[clap(long, env, default_value = "0")]
     pub starting_block: u64,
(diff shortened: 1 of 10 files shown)


src/contracts/batching/mod.rs
```diff
@@ -57,27 +54,31 @@ impl IdentityManager for Contract {
...
-        todo!()
+        self.tree_depth
     }
 
     fn initial_leaf_value(&self) -> Field {
-        todo!()
+        self.initial_leaf_value
     }
 
     fn group_id(&self) -> U256 {
```
kustosz, line 78: Just remove this altogether

src/prover/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// # Arguments
+    /// - `start_index`: The index in the merkle tree at which the insertions
+    ///   were started.
+    /// - `pre_root`: The value of the merkle tree's root before identities were
+    ///   inserted.
+    /// - `post_root`: The value of the merkle tree's root after the identities
+    ///   were inserted.
+    /// - `identities`: A list of identity insertions, ordered in the order the
+    ///   identities were inserted into the merkle tree.
+    pub async fn generate_proof(
```
kustosz, line 71: This should probably deal with batch sizes too?

src/prover/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+    /// The number of seconds to wait before timing out the transaction.
+    #[clap(long, env, default_value = "30")]
+    pub mtb_prover_timeout_secs: u64,
+}
+
+/// A representation of the connection to the MTB prover service.
+#[derive(Clone, Debug)]
+pub struct Prover {
+    target_url: Url,
+    client:     Client,
```
kustosz, line 39: please use `reqwest::Client` qualified, `Client` means nothing

src/app.rs
```diff
@@ -59,11 +63,14 @@ pub struct Options {
     pub ethereum: ethereum::Options,
 
     #[clap(flatten)]
-    pub contracts: contracts::Options,
+    pub identity_manager: contracts::Options,
```
kustosz, line 66: I'm not sure about this rename. From the perspective of a deployer, this deals with the contract. Altogether, "identity manager" is a nebulous term that means much less than "contract" but what it actually means to mean is "contract".

kustosz (approved):
Good stuff, some comments inline

src/app.rs
kustosz: Also it was rolled back by Greg because it broke existing deployments

src/contracts/batching/mod.rs
iamrecursion: It'll be removed in the next PR where I pull out the legacy contract entirely.

PR #386, opened 2023-02-10
Remove the legacy contract and integrate the batching one

### LINKED ISSUE #259: Batch insertion
Requires #258.

Modify the transaction worker service (B). From it's #258 behaviour:

1. It takes an item from the work-queue.
2. It flags it as in-progress.
2. It creates and signs an Ethereum transaction.
3. It waits for the transaction to be mined.
4. Marks the work item as done.

Into the following:

---

1. Check the work-queue. If it has more than `MAX_BATCH` unprocessed items or the oldest unprocessed item is older than `MAX_DELAY`, proceed.
2. Take up to `MAX_BATCH` unprocessed items from work queue, oldest first.
3. Mark these as 'processing' with a timestamp.
4. Request next available nonce for the transaction, and flag this nonce as taken.
5. Send the batch to a prover.
6. When the proof is done, create and sign a transaction.
7. Send transaction to mempool using one or more API endpoints.
8. Babysit the transaction, re-send and/or raise gasprice if necessary.
9. Once the transaction is mined and sufficiently confirmed, mark work items as done.

TBD:
* Create a ZKP prover services for the compute-heavy proving tasks
* Mechanism for recovering from failed work. For example a timeout based on the timestamp in 3.
* Do we want to split transaction signing/sending to a separate service? It can be a reusable component.

iamrecursion:
I am _so_ sorry for the size of this PR, but most of that horrendous line count is the new compiled contracts required for testing. It's big regardless, though.

This PR removes any use of the legacy contract from the codebase. This means that, while the code still exists due to some dependencies, it is no longer used. The batching contract is no-longer gated behind a feature flag.

It is now fully integrated with the batching contract on chain, as well as the prover service used to generate proofs for verification on chain. These interactions are tested, and while multiple components are mocked the signup sequencer portion of the code is as it should be with the real external dependencies.

Note that running the tests on macOS may trigger firewall protections that you need to disable. This depends on your macOS version.

Closes #259


kustosz (approved):
Looks good in general! Few nitpicks inline.

PR #387, opened 2023-02-11
/verifyProof endpoint

### LINKED ISSUE #401: Log mined roots
Currently we have no observability into what the last mined root is. We should log this such that we can at least find out via datadog.

### LINKED ISSUE #402: Record the root history in the database
Given that the signup sequencer is moving to be the source of truth for identity operations within a given group, we should also maintain the root history in the database.

This will involve creating a new table that contains the following:

- The root value.
- The status (pending or mined).
- The optional block timestamp for the block it was mined in.
- References to the first identity and last identity in the identities table that the batch contained.

From there, it just remains to connect it up so that the data gets recorded in the database.

gswirski:
## To Do
- [x] check root history
- [x] invalid root test (unknown or stale)
- [x] invalid proof test

Closes #402 
Closes #401


schemas/database/002_create_root_history.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE root_history
```
kustosz, line 1: this should also track root status w.r.t. chain status (i.e. is it even a mined-batch root and, in the future, is it confirmed)

src/app.rs
```diff
@@ -37,6 +38,27 @@ impl ToResponseCode for InclusionProofResponse {
     }
 }
 
+#[derive(Serialize)]
+pub enum VerifyProofResponse {
+    Ok,
```
kustosz, line 43: I think it would be better to just let the clients decide root expiration. I.e. provide them with root timestamp, rather than `Ok` vs `Expired`.

src/app.rs
```diff
@@ -62,16 +84,21 @@ pub struct Options {
     /// Timeout for the tree lock (seconds).
     #[clap(long, env, default_value = "120")]
     pub lock_timeout: u64,
+
+    /// Timeout for root validity (seconds).
+    #[clap(long, env, value_parser=duration_from_str, default_value="3600")]
+    pub root_validity_timeout: Duration,
```
kustosz, line 90: Yeah, I say let the caller decide.

src/database/mod.rs
```diff
@@ -243,6 +244,29 @@ impl Database {
...
+            sqlx::query(r#"SELECT seen_at FROM root_history WHERE root = $1 LIMIT 1;"#).bind(root);
+        let row = self.pool.fetch_optional(query).await?;
+        Ok(row.map(|r| r.get::<_, _>(0)))
+    }
+
+    pub async fn store_root_state(&self, root: &Hash) -> Result<(), Error> {
+        let query = sqlx::query(
+            r#"INSERT INTO root_history(root, seen_at)
+                    VALUES ($1,  CURRENT_TIMESTAMP)
+                    ON CONFLICT (root)
```
kustosz, line 262: This should `ON CONFLICT DO NOTHING`, makes no domain sense to bump the root timeout here.

src/app.rs
```diff
@@ -245,6 +276,36 @@ impl App {
...
+    #[instrument(level = "debug", skip_all)]
+    pub async fn verify_proof(
+        &self,
+        request: &VerifyProofRequest,
+    ) -> Result<VerifyProofResponse, ServerError> {
+        let Some(root_timestamp) = self.database.get_root_timestamp(&request.root).await? else {
+            return Ok(VerifyProofResponse::InvalidRoot)
+        };
+
+        if root_timestamp + self.root_validity_timeout < Utc::now() {
```
kustosz, line 291: This is too simplistic. Suppose inserts stopped for 12 hrs. Then the latest root will be deemed invalid. We've already changed this in the contract such that:
1. Latest root is always valid.
2. Root expiration starts counting when the root is superseded, not when it's created.

Essentially, if we stop inserts for 12 hrs, then this long-lived root should not become invalid as soon as next insert comes, but rather it should remain valid for a set period.

tests/integration_tests.rs
```diff
@@ -408,12 +594,33 @@ fn construct_inclusion_proof_body(identity_commitment: &Hash) -> Body {
...
-fn construct_insert_identity_body(identity_commitment: &str) -> Body {
-    Body::from(
+fn construct_insert_identity_body(identity_commitment: &Field) -> Body {
+    let body = Body::from(
         json!({
             "identityCommitment": identity_commitment,
         })
         .to_string(),
+    );
+    println!("body {:?}", body);
```
kustosz, line 604: this is an oopsie

kustosz (changes requested):
This code is subtly incorrect, explained inline.

src/server.rs
```diff
@@ -64,6 +65,17 @@ pub struct InclusionProofRequest {
     pub identity_commitment: Hash,
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[serde(deny_unknown_fields)]
+pub struct VerifyProofRequest {
```
iamrecursion, line 71: It is definitely worth documenting that this endpoint corresponds to the verification of _semaphore_ protocol proofs, and not the merkle tree (insertion/update/removal) proofs. Probably worth renaming that to be clear.

iamrecursion (changes requested):

src/app.rs
gswirski: Removed.

src/app.rs
gswirski: Fixed.

iamrecursion (approved):
LGTM. Just please fix the clippy lints.

src/app.rs
```diff
@@ -37,6 +39,24 @@ impl ToResponseCode for InclusionProofResponse {
     }
 }
 
+#[derive(Serialize)]
+pub enum VerifySemaphoreProofResponse {
```
Dzejkop, line 43: Sorry to be that guy but can we talk a bit about the format of this response? I find that serde's default enum serialization is kind of ugly, the `Ok` response serializes to:

```rust
{
    "Ok": {
        "root": "0x0000000000000000000000000000000000000000000000000000000000000000",
        "status": "pending",
        "pending_valid_as_of": "2023-03-07T19:51:01Z",
        "mined_valid_as_of": null
    }
}
```

whereas any of the error responses serializes to a string.

If we go with something like
```rust
#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
pub enum VerifySemaphoreProofResponse {
    Ok(RootItem),
    Err(VerifySemaphoreProofResponseError),
}

#[derive(Serialize)]
#[serde(rename_all = "camelCase")]
pub enum VerifySemaphoreProofResponseError {
    InvalidRoot,
    InvalidProof,
    ProverError,
}
```

then it'll serialize to either
```json
{
    "ok": {
        "root": "0x0000000000000000000000000000000000000000000000000000000000000000",
        "status": "pending",
        "pending_valid_as_of": "2023-03-07T19:51:01Z",
        "mined_valid_as_of": null
    }
}
```

or

```json
{
    "err": "invalidRoot"
}
```

src/app.rs
```diff
@@ -245,6 +270,32 @@ impl App {
...
+            request.root,
+            request.nullifier_hash,
+            request.signal_hash,
+            request.external_nullifier_hash,
+            &request.proof,
+        );
+        match checked {
+            Ok(true) => Ok(VerifySemaphoreProofResponse::Ok(root_state)),
+            Ok(false) => Ok(VerifySemaphoreProofResponse::InvalidProof),
+            Err(_) => Err(ServerError::ProverError),
```
Dzejkop, line 295: Can we log the error here for visibility? It could happend pretty often so maybe on `debug` level?

src/database/mod.rs
```diff
@@ -243,10 +253,167 @@ impl Database {
...
 
 #[derive(Debug, Error)]
 pub enum Error {
-    #[error("database error")]
+    #[error("database error: {0}")]
     InternalError(#[from] sqlx::Error),
 }
+
+#[cfg(test)]
+#[allow(unused_imports)]
```
Dzejkop, line 333:
```suggestion
```

src/database/mod.rs
```diff
@@ -243,10 +253,167 @@ impl Database {
...
+        &self,
+        root: &Hash,
+        last_identity: &Hash,
+        last_index: usize,
+    ) -> Result<(), Error> {
+        let query = sqlx::query(
+            r#"INSERT INTO
+            root_history(root, last_identity, last_index, status, created_at)
+            VALUES($1, $2, $3, $4, STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW'))
+            ON CONFLICT (root) DO NOTHING;"#,
```
Dzejkop, line 314: If there's a conflict, would that potentially mean that we're syncing our internal invalid state to blockchain valid state and therefore we should mark this as pending again?

I know you mentioned that Database is now the source of truth, so maybe it's fine, just want to make sure.

Dzejkop (approved):
Left some minor stylistic comments (mostly about response format), otherwise LGTM

src/app.rs
gswirski: I made `VerifySemaphoreProofResponse` transparent and moved errors to the parent result object.

kustosz (approved):

PR #392, opened 2023-02-14
feat: deployment for second instance of signup-sequencer

jazzbee:
Second deployment of signup-sequencer as signup-sequencer-v2 with additional workflow to automatically deploy everything merged into batching/main.Signup-sequncer-v2  is available at signup-v2.stage-crypto.worldcoin.dev


.github/workflows/build-test-deploy.yml
```diff
@@ -337,6 +337,38 @@ jobs:
           --set env.SEMAPHORE_ADDRESS="${{ secrets.SEMAPHORE_CONTRACT_ADDRESS }}"
           -f deploy/main-stage.values.yaml
 
+  deploy-batching-main-stage:
+    name: Deploy main to staging
```
gswirski, line 341: rename to "Deploy batching/main to batching staging"?

gswirski (changes requested):
I don't see `cargo build` nor `cargo run` anywhere in this config. We need to run with `--features oz-provider` flag (so something like `cargo run --features oz-provider`)

gswirski (approved):

Issue #363, opened 2023-01-24
Properly use `world-id-contracts` as a dependency


iamrecursion:
Currently we hard-code the ABIs of the contracts with which we are working. This is _very_ brittle in the face of changes to the contract ABIs. Instead, in the future, we want to have the `build.rs` compile and evaluate the ABIs for the contracts directly using solidity.

- [ ] Make it grab `world-id-contracts` as a dependency.
- [ ] Make the `build.rs` execute the solidity compiler and then generate ABI definitions for `ethers-rs` from the resultant artefacts. 
- [ ] Make it possible to detect the latest version of the batching contract in a given commit (by contract name).
- [ ] Have it always ensure that it uses the ABI for the latest version of the identity manager.

PR #395, opened 2023-02-16
Update the Contract ABI

iamrecursion:
The contract ABIs and compiled contract files used for testing had become out of date due to changes in WorldID contracts.

This PR incorporates the changes from https://github.com/worldcoin/world-id-contracts/pull/51, but does not depend directly on it. All tests still run.


kustosz (approved):
wow we really need to automate this bit :P

iamrecursion:
Yeah. Not like I filed #363 or anything~

PR #398, opened 2023-02-23
Have the Prover Mock Verify Inputs

### LINKED ISSUE #391: Verify Inputs in the Prover Service Mock
Currently the mock prover service used in the integration tests does no inspection of its input at all. This makes it possible to sneak through subtle logic bugs. 

Instead, we want to actually verify the inputs in the mock prover service, allowing us to ensure that things make sense. This should be done as follows:

- [x] Replicate the logic of the [circuit](https://github.com/worldcoin/semaphore-mtb/blob/main/prover/circuit.go#L80) from the prover without it being a ZK circuit.
- [x] If it fails to prove, things are invalid, so return an error response.
- [x] If it succeeds, return a response that is correlated with the input (e.g. `a[0] % 2 == 0 && a[1] == inputHash`).
- [x] Create a mock verifier in solidity that verifies based on _this_ condition instead.
- [x] Use that verifier in the integration tests for signup sequencer.

iamrecursion:
This commit ensures that the mock prover service now actually performs verification on the provided proofs. This means that only a malicious signup sequencer should be able to spuriously pass the integration tests.

In addition, it ensures that the return value from the prover mock (the proof) is now correlated with the input in a way that can be verified on the mock chain.

Corresponds with but does not depend on https://github.com/worldcoin/world-id-contracts/pull/52. 

Closes #391.


tests/prover_mock/mod.rs
```diff
@@ -100,7 +116,119 @@ impl Service {
...
+    /// We keccak hash all input to save verification gas. Inputs are arranged
+    /// as follows:
+    /// ```
+    /// StartIndex || PreRoot || PostRoot || IdComms[0] || IdComms[1] || ... || IdComms[batchSize-1]
+    ///     32     ||   256   ||   256    ||    256     ||    256     || ... ||     256 bits
+    /// ```
+    fn calculate_identity_registration_input_hash(input: &ProofInput) -> U256 {
+        // Calculate the input hash as described by the prover.
+        let mut hashable_bytes: Vec<u8> = vec![];
+        let mut reserved_u256_bytes: [u8; size_of::<U256>()] = Default::default();
```
gswirski, line 220: nit: I'd rename to something like `buffer` to indicate that it doesn't really do anything on its own. Looking at the name, I was expecting some special reserved block in input bytes.

tests/prover_mock/mod.rs
```diff
@@ -100,7 +116,119 @@ impl Service {
...
     }
+
+    /// Performs the 'proof' operation on the provided `input`.
+    ///
+    /// Note that this does _not_ implement the full ZK proof system as done by
+    /// `semaphore-mtb`. Instead, it just verifies that the provided merkle
+    /// proofs are correct, which is sufficient when combined with the mock
+    /// verifier used by the mock chain.
+    ///
+    /// In order to safe effort and reduce the surface for bugs, the proof
```
gswirski, line 131: nit: safe -> save

gswirski (approved):
Looking great! Minor irrelevant nits.

PR #405, opened 2023-02-28
Prometheus pending identities count

cichaczem:
## Motivation

This is one of the multiple small PRs adding system statistics to Prometheus. As a first one, we count pending identities (not submitted to the chain yet)

## Solution

The pending identities count comes from the database, the main source of truth. For starters, we can set this value every tick or when triggered by request.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/identity_committer.rs
```diff
@@ -3,9 +3,11 @@ use std::{
     time::{Duration, SystemTime},
 };
 
+use ::prometheus::{register_gauge, Gauge};
```
gswirski, line 6: nit: do we need leading `::`? Looks suspicious.

src/identity_committer.rs
```diff
@@ -379,6 +388,14 @@ impl IdentityCommitter {
             .await?;
         mined_tree.apply_next_updates(updates.len()).await;
 
+        Self::log_pending_identities_count(database).await?;
```
gswirski, line 391: I think it's more a question than actual feedback. What would be the best practice in Prometheus?
* Update metrics with specific frequency, away from the code that manipulates data; or
* Update metrics along the code that manipulates underlaying data

The current implementation is nice because it captures the full history of all "pending count" data points. However, if metrics collector only fetches snapshots every now and then, we will lose that history later in the pipeline.

On the other hand, we will have more statuses than just `pending`. Maybe it makes sense to have a completely separate task that periodically checks `identities` table and updates all related metrics? This way, as the codebase grows more complicated, we don't forget to update metrics each time we manipulate the data.

gswirski (approved):

src/identity_committer.rs
cichaczem: I thought about creating a separate task too. I wanted to defer this decision until we have a full view of the stats we want to collect. I think we're getting close so let's use this solution and extract it when we work on the next batch of stats

PR #407, opened 2023-03-06
Refactor OpenZeppelin Relayer integration

Dzejkop:


src/contracts/mod.rs
```diff
@@ -123,7 +124,7 @@ impl IdentityManager {
     }
 
     #[instrument(level = "debug", skip_all)]
-    pub async fn is_owner(&self) -> anyhow::Result<bool> {
+    pub async fn _is_owner(&self) -> anyhow::Result<bool> {
```
iamrecursion, line 127: Why?

src/contracts/mod.rs
```diff
@@ -205,7 +206,7 @@ impl IdentityManager {
     }
 
     #[instrument(level = "debug", skip_all)]
-    pub async fn assert_valid_root(&self, root: Field) -> anyhow::Result<()> {
+    pub async fn _assert_valid_root(&self, root: Field) -> anyhow::Result<()> {
```
iamrecursion, line 209: Why?

src/contracts/mod.rs
Dzejkop: They were unused (with and without `oz-provider` feature) which was throwing compilation warnings

gswirski: Let's ✂️ it then. :)

src/contracts/mod.rs
gswirski: Let's ✂️ it then.

gswirski (approved):
👍

PR #410, opened 2023-03-07
Poll oz transactions

Dzejkop:
## Motivation
It can take a long time to confirm a transaction on the chain. In that time we could have been already preparing and proving the next batch.

## Solution
"Sending" an identities commitment transactions is split into 3 phases:
1. Preparing the proof - prepares proof data and sends it to the prover
2. Sending the transaction - only sends the transaction to the RPC (or OZ Defender Relay)
3. Mining the transaction - awaiting for the transaction to be finalized on chain

The process is synchronized with a channel in such a way that we can already start preparing the next proof while the previous transaction is still unconfirmed.

## Also in this PR1

1. Added `total_proving_time` and `prover_proving_time` metrics

## TODO:
1. [x] ~Tests~ - will work on testing later, codebase needs a bit more work to enable that
2. [ ] Display added metrics on a dashboard

## PR Checklist

- [x] ~Added Tests~
- [x] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ ethers = { version = "1.0.0", features = [ "ws", "ipc", "openssl", "abigen" ] }
 eyre = "0.6"
 futures = "0.3"
 futures-util = { version = "^0.3" }
+hex = "0.4.3"
 hyper = { version = "^0.14.17", features = ["server", "tcp", "http1", "http2"] }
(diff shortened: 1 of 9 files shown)


iamrecursion (approved):
Please fix (or suppress if there's no sensible fix) the clippy lints, but otherwise LGTM.

PR #416, opened 2023-03-13
Batching/oz api crate

Dzejkop:
## Solution
Splits the OpenZeppelin Defender Relay integration code into a separate crate - better separation of concerns, possible compile time improvements.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/crates/oz-api/src/auth.rs
+++ b/crates/oz-api/src/auth.rs
@@ -1,6 +1,7 @@
-use anyhow::{Context, Result as AnyhowResult};
 use cognitoauth::cognito_srp_auth::{auth, CognitoAuthInput};
 use hyper::{http::HeaderValue, HeaderMap};
+
+use crate::error::Error;
 
 // Same for every project, taken from here: https://docs.openzeppelin.com/defender/api-auth
 const CLIENT_ID: &str = "1bpd19lcr33qvg5cr3oi79rdap";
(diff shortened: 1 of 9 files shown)


crates/oz-api/src/auth.rs
```diff
@@ -0,0 +1,50 @@
...
+
+// Same for every project, taken from here: https://docs.openzeppelin.com/defender/api-auth
+const CLIENT_ID: &str = "1bpd19lcr33qvg5cr3oi79rdap";
+const POOL_ID: &str = "us-west-2_iLmIggsiy";
+
+#[derive(Clone, Debug)]
+pub struct ExpiringHeaders {
+    pub headers:         HeaderMap,
+    /// The timestamp at which the headers will expire in seconds
+    pub expiration_time: i64,
```
iamrecursion, line 14: Why use a raw `i64` over some kind of wrapped time type? If the `i64` makes the most sense, I'd consider either renaming to `experiation_time_secs` or wrapping it in a newtype `ExpirationTimeSeconds` for clarity when not staring at this section of the code.

iamrecursion (approved):
Great stuff! Nicely architected and clean. Just one nit inline.

crates/oz-api/src/auth.rs
Dzejkop: I basically copied this from somewhere else and in that code we got this value from an 'expiresIn' field from the response. I'll see about changing it to 'Instant' or something from chrono

iamrecursion (approved):
Great stuff!

PR #422, opened 2023-03-14
Batching/split tasks into files

Dzejkop:
Small PR that splits up the `identity_commiter.rs` file into multiple - one for each task.


src/identity_committer/tasks/process_identities.rs
```diff
@@ -0,0 +1,282 @@
...
+                    // or the timer ticks.
+                    timer.reset();
+                    last_batch_time = SystemTime::now();
+                }
+            }
+        }
+    }
+
+    // TODO This can be split into multiple phases. The first would compute the
+    //   batches and the second would mine those batches.
```
iamrecursion, line 136: Is this `TODO` still relevant?

iamrecursion (approved):
Great stuff. One question inline.

src/identity_committer/tasks/process_identities.rs
Dzejkop: It's not, removed it

PR #432, opened 2023-03-21
Memory efficient and type safe merkle trees

kustosz:
Integrates the half-lazy trees (allowing us to run with d=30) and makes versions API typesafe.


src/identity_tree.rs
```diff
@@ -28,181 +34,340 @@ impl TreeUpdate {
...
+    }
+
+    /// Garbage collection for the canonical tree version. It rewrites all
+    /// future versions of the tree to use the more optimized storage of this
+    /// tree. This is done periodically, to really make the additional
+    /// time-costs of these rebuilds repay themselves in saved memory.
+    ///
+    /// Warning: this will attempt to acquire lock for all transitive successors
+    /// of this tree, and therefore no version locks acquired through
+    /// `TreeVersion#get_data()` may be held at the time of calling this.
```
iamrecursion, line 214: We likely can't do better in the type system, but it might be good to try to lock the lock and then raise an error before backing out or dying.

src/identity_tree.rs
```diff
@@ -28,181 +34,340 @@ impl TreeUpdate {
...
+                next.get_data().rebuild_on(self.tree.derived());
             }
         }
-        proofs
     }
+}
+
+impl TreeVersionData<lazy_merkle_tree::Derived> {
+    /// Reapplies all changes of the given tree. This is only used for garbage
+    /// collection – `tree` will usually be a more densely stored version of the
```
iamrecursion, line 228: Usually? Under what circumstances would it not be? When it's equally dense?

src/utils.rs
```diff
@@ -72,3 +72,11 @@ where
...
     value.to_string().parse::<f64>().unwrap()
 }
+
+pub fn replace_with<T>(value: &mut T, modifier: impl FnOnce(T) -> T) {
+    unsafe {
+        let v = ptr::read(value);
+        let v = modifier(v);
+        ptr::write(value, v);
+    }
+}
```
iamrecursion, line 82: Doc it.

iamrecursion (approved):
Unfortunately I can't think of a way to hide what you want to be hidden without a totally different architecture, but this looks really good. I know I'm gonna appreciate this when I come to work with this code again.

Please fix the clippy lints and make sure to `cargo fmt --all`.

src/identity_tree.rs
kustosz: This lock can be congested due to legitimate reasons, this comment is just so that we don't deadlock internally.

PR #434, opened 2023-03-22
Send batch sizes to Prometheus

cichaczem:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/src/identity_committer/tasks/process_identities.rs
+++ b/src/identity_committer/tasks/process_identities.rs
@@ -74,8 +74,6 @@ impl IdentityCommitter {
                         &updates
                     ).await?;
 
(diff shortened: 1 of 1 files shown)


src/identity_committer/tasks/process_identities.rs
```diff
@@ -74,6 +74,8 @@ impl IdentityCommitter {
                         &updates
                     ).await?;
 
+                    Self::log_batch_size(updates.len()).await?;;
```
Dzejkop, line 77: We could move this inside of `commit_identitites` and then we wouldn't need to have this kind of duplication.

Dzejkop (approved):
One small comment, but LGTM

PR #436, opened 2023-03-23
Batching/service resillience

Dzejkop:
This PR:
1. Removes panics and `.unwrap`s from a few places (sometimes `.unwrap`s are replaced with `.expect`s)
2. Adds an outer loop to both processing tasks to ensure they'll restart even if they panic
3. Minor formatting & clippy demanded changes
4. Makes the integration tests run in parallel - all the ports are acquired from the operating system - so there should be no clashes
5. Minor refactoring of integration tests - notably a speed up of mock chain deployment + setting up dependencies (db, chain, prover) in parallel
6. Splits up integration tests into separate files - so that they run in separate processes

TODO: Since integration tests are now separate processes - we should be able to run them in parallel using [Nextest](https://nexte.st/)


src/identity_committer.rs
```diff
@@ -135,62 +138,98 @@ impl IdentityCommitter {
...
-                        timeout
-                    ) => {
-                        result?;
-                    }
-                    _ = shutdown_receiver.recv() => {
-                        info!("Woke up by shutdown signal, exiting.");
-                        return Ok(());
+            crate::utils::spawn_monitored_with_backoff(
+                move || {
+                    let wake_up_receiver = wake_up_receiver.clone();
```
Dzejkop, line 158: I'm not happy with the second level of clones to move the data into the context - and there definitely is a refactor needed here but I want to move on to the db unification + actually setting up the Task Failed monitoring

tests/insert_identity_and_proofs.rs
```diff
@@ -0,0 +1,239 @@
+mod common;
+
+use common::prelude::*;
+
+#[tokio::test]
+#[serial_test::serial]
+async fn insert_identity_and_proofs() -> anyhow::Result<()> {
```
Dzejkop, line 7: Nothing changed in this tests - it's just split

tests/unavailable_prover.rs
```diff
@@ -0,0 +1,136 @@
+mod common;
+
+use common::prelude::*;
+
+/// Tests that the app can keep running even if the prover returns 500s
+/// and that it will eventually succeed if the prover becomes available again.
+#[tokio::test]
+async fn unavailable_prover() -> anyhow::Result<()> {
```
Dzejkop, line 8: This test is actually new

tests/common/prover_mock.rs
```diff
@@ -109,34 +136,47 @@ impl Service {
...
+        };
+
         Ok(service)
     }
 
+    pub fn url(&self) -> String {
+        format!("http://{}", self.address)
+    }
+
+    pub async fn set_availability(&self, availability: bool) {
```
Dzejkop, line 157: Added an option to make the prover return 500s for every request

src/utils.rs
```diff
@@ -42,33 +34,146 @@ where
     }
 }
 
-/// Spawn a task and abort process if it panics or results in error.
-pub fn spawn_or_abort<F, T>(future: F) -> JoinHandle<T>
+pub fn spawn_monitored_with_backoff<S, F, T>(
```
Dzejkop, line 37: This handles the spawning and respawning of task futures, while waiting after every failure

cichaczem (approved):

src/utils.rs
```diff
@@ -42,33 +34,146 @@ where
...
+            // Wrap in `AssertUnwindSafe` so we can call `FuturesExt::catch_unwind` on it.
+            let future = std::panic::AssertUnwindSafe(future);
+
+            let result = future.catch_unwind().await;
+
+            match result {
+                // Task succeeded or is shutting down gracefully
+                Ok(Ok(t)) => return t,
+                Ok(Err(e)) => {
+                    error!("Task failed: {:?}", e);
```
gswirski, line 60: very minor nit, surprised that clippy doesn't catch this:
`error!("Task failed: {e:?}");`

src/utils.rs
```diff
@@ -42,33 +34,146 @@ where
...
+
+                            panic!("Panicking!");
+                        }
+                    }
+                },
+                Duration::from_secs_f32(0.2),
+            )
+        };
+
+        println!("Sleeping for 1 second");
```
gswirski, line 158: `info!`?

gswirski (approved):
Looks like a solid upgrade! 💪

I flagged two very very minor nits. We should merge this to unblock your db work, and can rethink double move later.

src/utils.rs
Dzejkop: Actually this is test output and I'm not spinning up a tracing subscriber, so that wouldn't work. I'll leave the printlns

Dzejkop:
[#457](https://github.com/worldcoin/signup-sequencer/issues/457)

PR #438, opened 2023-03-23
Update schema for verifySemaphoreProof

Dzejkop:


schemas/openapi.yaml
```diff
@@ -66,6 +66,38 @@ paths:
...
+              schema:
+                type: string
+                example: 'invalid root'
+        '500':
+          description: Prover error
+          content:
+            text/plain:
+              schema:
+                type: string
+                example: 'prover errorroot'
```
gswirski, line 99: minor typo. -> `prover error`

gswirski (approved):

schemas/openapi.yaml
Dzejkop: Good catch, thanks!

iamrecursion (approved):

PR #440, opened 2023-03-24
Batching/unify db

Dzejkop:
## Motivation
To deduplicate the status value held in the database and to simplify some queries while also more strongly ensuring the one-to-one mapping of identity to root hash.

Also added quite a few more granular database tests.

In the end not that much has changed - we still have the split between the identities and root_history column, because in the app we store the identity first and then we store the root hash after updating the tree.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog


src/database/mod.rs
```diff
@@ -121,113 +125,103 @@ impl Database {
...
+        };
+
+        let root_leaf_index = root_leaf_index as i64;
+
         let update_root_history_query = sqlx::query(
             r#"
             UPDATE root_history
-            SET status = $1, mined_at = CURRENT_TIMESTAMP
-            WHERE last_index <= $2
+            SET status = 'mined', mined_at = CURRENT_TIMESTAMP
```
gswirski, line 216: Use `Status::Mined` to avoid magic literals

src/app.rs
```diff
@@ -127,7 +127,11 @@ impl App {
         let root_hash = identity_manager.latest_root().await?;
         let root_hash = root_hash.into();
 
-        database.mark_root_submitted_to_contract(&root_hash).await?;
+        // We don't store the initial root in the database, so we have to skip this step
```
kustosz, line 130: Not really something we have to do right now, but it would be good to include a healthcheck for "does every identity have a root" somewhere on startup. Maybe even a separate monitor thread that checks that periodically and sounds an alarm if this happens? Even trying to insert these roots retroactively. This problem can easily arise if we have DB connection issues and this is an important stability assumption.

src/database/mod.rs
```diff
@@ -1,4 +1,8 @@
-#![allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
+#![allow(
+    clippy::cast_sign_loss,
+    clippy::cast_possible_truncation,
+    clippy::needless_range_loop
```
kustosz, line 4: I agree with clippy here, there are needless range loops in this file and its suggestion is better style.

src/database/mod.rs
```diff
@@ -393,21 +367,41 @@ impl Database {
...
 
     use super::{Database, Options};
     use crate::identity_tree::Status;
 
-    #[tokio::test]
-    async fn test_root_invalidation() -> anyhow::Result<()> {
+    macro_rules! assert_same_time {
+        ($a:expr, $b:expr, $diff:expr) => {
+            assert!(
+                $a - $b < $diff,
```
kustosz, line 391: `chrono` allows negative durations, so you probably also want to check `$b -$a < $diff`. In fact, I think most uses of this macro in these tests are vacuous, in that `$a` is usually before `$b`.

src/database/mod.rs
```diff
@@ -121,113 +125,105 @@ impl Database {
         Ok(Self { pool })
     }
 
+    pub async fn is_empty(&self) -> Result<bool, Error> {
```
kustosz, line 128: Should be called `has_no_identities` or something similar. Currently calling `database.is_empty()` makes sense, but if we add more tables in the future this name will be misleading (but will probably not be changed at that point because the future versions of us will lack this context).

kustosz (approved):
Great stuff in general. There are a few minor issues pointed out inline. There is also one low probability but very high impact issue, where a well-timed infra failure will lead to the app never being able to sync with contract. This would be pretty bad, but I know we are in a hurry to use the new DB schema, so I'm approving this, but that improvement to startup logic should be a high-priority fix.

kustosz (approved):
Very nice and much safer than before! Ship it!

PR #449, opened 2023-04-03
Batching Support for multiple batch sizes + bug fix

Dzejkop:
Changes:
1. The sequencer no longer accepts the following cli args `--batch-size`, `--batch-insertion-prover-timeout` or `--batch-insertion-prover-url`,
    instead it accepts `--prover-urls` which accepts a JSON list of the following format: `[{"url": "http://localhost:3001","batch_size": 3,"timeout_s": 30}]`
2. If sending a batch which is not full - the sequencer will choose the smallest batch size out of the available provers that can contain the entire batch
3. Fixes a bug that would leave the sequencer in an invalid state if the `process_identitites` task failed or panicked

### DIFF
--- a/src/identity_tree.rs
+++ b/src/identity_tree.rs
@@ -6,7 +6,7 @@ use std::{
 use chrono::Utc;
 use semaphore::{
     lazy_merkle_tree,
-    lazy_merkle_tree::LazyMerkleTree,
+    lazy_merkle_tree::{Derived, LazyMerkleTree},
     merkle_tree::Hasher,
     poseidon_tree::{PoseidonHash, Proof},
     Field,
@@ -99,7 +99,13 @@ pub struct CanonicalTreeMetadata {
 /// Additional data held by any derived tree version. Includes the list of
 /// updates performed since previous version.
 pub struct DerivedTreeMetadata {
-    diff: Vec<TreeUpdate>,
+    diff: Vec<TreeUpdateWithTree>,
+}
+
+#[derive(Clone)]
+pub struct TreeUpdateWithTree {
+    pub update: TreeUpdate,
+    pub tree:   PoseidonTree<Derived>,
 }
 
 /// Trait used to associate a version marker with its metadata type.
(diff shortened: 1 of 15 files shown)


src/identity_tree.rs
```diff
@@ -156,48 +162,48 @@ where
...
+
+            let updates: Vec<&TreeUpdate> = next
+                .metadata
+                .diff
+                .iter()
+                .map(|u| &u.update)
+                .take(update_count)
+                .collect();
+
+            for update in &updates {
```
kustosz, line 194: Probably could be smarter about this – no need to recompute diffs and result, could just reuse from the applied updates. We could add something like `apply_diffs(&mut self, diff: Vec<TreeUpdateWithTree>)` to `BasicTreeOps`

src/prover/mod.rs
```diff
@@ -1,12 +1,94 @@
...
+
+        Ok(Self(map))
+    }
+}
+
+impl<P> ProverMap<P> {
+    /// Get the smallest prover that can handle the given batch size.
+    pub fn get(&self, batch_size: usize) -> Option<&P> {
+        self.0.iter().find_map(|(size, prover)| {
+            if batch_size <= *size {
```
kustosz, line 49: this whole if is same as `(batch_size <= *size).then_some(prover)`

src/task_monitor/tasks/process_identities.rs
```diff
@@ -201,27 +201,44 @@ async fn commit_identities(
...
     let pre_root: U256 = batching_tree.get_root().into();
-    let mut commitments: Vec<U256> = updates.iter().map(|update| update.element.into()).collect();
+    let mut commitments: Vec<U256> = updates
+        .iter()
+        .map(|update| update.update.element.into())
+        .collect();
+
+    let latest_tree_from_updates = updates.last().expect("Updates is non empty.").tree.clone();
 
     // Next we apply the updates, retrieving the merkle proofs after each step of
```
kustosz, line 229: this comment aged like milk, please fix

kustosz (approved):
A few minor nitpicks, but great stuff overall

src/task_monitor/tasks/process_identities.rs
```diff
@@ -201,27 +201,44 @@ async fn commit_identities(
...
     // that process.
-    let mut merkle_proofs = batching_tree.apply_next_updates(updates.len());
+    // TODO: Change to operator on an immutable version of the tree
```
iamrecursion, line 231:
```suggestion
    // TODO: Change to operate on an immutable version of the tree
```

Why?

iamrecursion (approved):
Really great stuff! A few clippy lints that should be fixed and a couple of comments inline, but other than that this lgtm.

src/identity_tree.rs
Dzejkop: done

PR #450, opened 2023-04-04
Replace wake up sender channel with notify

Dzejkop:
Replaces the wake up channel with a `Notify`. So that we are not limited to the number of identities inserted as pending into the database.

### DIFF
--- a/src/task_monitor.rs
+++ b/src/task_monitor.rs
@@ -6,7 +6,7 @@ use ethers::types::U256;
 use once_cell::sync::Lazy;
 use prometheus::{linear_buckets, register_gauge, register_histogram, Gauge, Histogram};
 use tokio::{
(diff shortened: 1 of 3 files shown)


src/task_monitor.rs
```diff
@@ -128,13 +128,13 @@ impl TaskMonitor {
         // We could use the second element of the tuple as `mut shutdown_receiver`,
         // but for symmetry's sake we create it for every task with `.subscribe()`
         let (shutdown_sender, _) = broadcast::channel(1);
-        let (wake_up_sender, wake_up_receiver) = mpsc::channel(1);
```
Dzejkop, line 131: I would also be partial to increasing the capacity of this channel, but I see no reason why we should limit the speed at which we insert pending identities into the database.

cichaczem (approved):
LGTM

PR #452, opened 2023-04-05
Minor bugfix

Dzejkop:


src/app.rs
```diff
@@ -122,7 +122,7 @@ impl App {
 
         // We don't store the initial root in the database, so we have to skip this step
         // if the db is empty
-        if !database.has_no_identities().await? {
+        if !database.has_no_mined_identities().await? {
```
gswirski, line 125: nit: if this is the only use of this function, wouldn't it be better to rename it to `has_mined_identities` and drop `!`?

PR #485, opened 2023-04-18
Add support for adding and removing batch sizes at runtime

### LINKED ISSUE #466: Support dynamic batch sizes
While the sequencer currently selects the correct batch size from those it was configured with at launch, we want the ability to add and remove batch sizes dynamically without having to restart the sequencer.

iamrecursion:
I've had some issues running `clippy` locally due to a failed download, and I expect the same to currently happen on CI.

Closes #466


src/server/mod.rs
```diff
@@ -221,6 +173,23 @@ async fn route(request: Request<Body>, app: Arc<App>) -> Result<Response<Body>,
             })
             .await
         }
+        (&Method::POST, "/addBatchSize") => {
```
Dzejkop, line 176: Would probably be useful to have an endpoint to list the current available batch sizes.

Also now I'm thinking it'd be great to have some fancier routing, maybe with `axum`? But no biggie, I can also take a stab at that in some spare moment.

src/server/mod.rs
```diff
@@ -1,11 +1,13 @@
+pub mod error;
```
Dzejkop, line 1: Love it! Thanks!

src/contracts/mod.rs
```diff
@@ -261,6 +271,48 @@ impl IdentityManager {
 
         Ok(latest_root)
     }
+
+    /// # Errors
+    ///
+    /// Will return `Err` if the provided batch size already exists.
+    pub async fn add_batch_size(
+        &self,
+        url: impl Into<String>,
```
Dzejkop, line 280: `impl ToString`?

tests/dynamic_batch_sizes.rs
```diff
@@ -0,0 +1,195 @@
...
+    let second_prover = spawn_mock_prover(second_batch_size).await?;
+    test_add_batch_size(&uri, second_prover.url(), second_batch_size as u64, &client)
+        .await
+        .expect("Failed to add batch size.");
+
+    // Insert enough identities to trigger the lower batch size.
+    test_insert_identity(&uri, &client, &mut ref_tree, &identities_ref, 3).await;
+    test_insert_identity(&uri, &client, &mut ref_tree, &identities_ref, 4).await;
+    tokio::time::pause();
+    tokio::time::resume();
```
Dzejkop, line 124: ?

tests/dynamic_batch_sizes.rs
```diff
@@ -0,0 +1,195 @@
...
+    .await;
+
+    // Add a new prover for batch sizes of two.
+    let second_prover = spawn_mock_prover(second_batch_size).await?;
+    test_add_batch_size(&uri, second_prover.url(), second_batch_size as u64, &client)
+        .await
+        .expect("Failed to add batch size.");
+
+    // Insert enough identities to trigger the lower batch size.
+    test_insert_identity(&uri, &client, &mut ref_tree, &identities_ref, 3).await;
```
Dzejkop, line 121: Could we maybe also make the larger prover unavaiable here?

With

```
prover_mock.set_availability(false).await;
```

Just to be sure that we're using the correct prover in this case.

gswirski (changes requested):
One big issue of this approach (and I'm sorry I haven't thought about this earlier) is that none of it is persistent. If the sequencer restarts (and it will e.g. for updates), we lose the entire state.

Two options here:
* a bit hacky: `/addBatchSize` and `/removeBatchSize` should return text saying something along the lines of: "don't forget to update PROVER_URLS to 'current prover map JSON'"
* `add_batch_size` and `remove_batch_size` functions update database state which takes precedence over PROVER_URLS. Then either rename PROVER_URLS to DEFAULT_PROVER_URLS, or delete it entirely and add setup instructions in the README (imho preferred approach_.

tests/dynamic_batch_sizes.rs
iamrecursion: Added an explanatory comment. It forces the timer to trigger.

src/server/mod.rs
iamrecursion: I've added the endpoint but more complex routes are out of scope for this PR.

iamrecursion:
Where this is left is with the feedback from Jakub addressed, but no work done with regards to persisting the state in the database. 

This made the most sense to me as a hand-over point as in my remaining time on the project for now I don't think I'd get further than part-way through the database work. It would likely be easier for somebody else to pick up that portion of the feature in its entirety rather than try and enter half-way through the implementation.

gswirski:
@TheMenko I merged your PR into this. Please resolve conflicts and you have my approval.
@cichaczem can you please double check if this is now good to deploy?

gswirski (approved):

cichaczem (approved):

PR #503, opened 2023-05-02
Add test workflow

Dzejkop:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/tests/common/chain_mock.rs
+++ b/tests/common/chain_mock.rs
@@ -139,13 +139,15 @@ pub async fn spawn_mock_chain(
         .clone()
         .deploy((first_batch_size as u64, mock_verifier.address()))?
         .confirmations(0usize)
-        .send();
+        .send()
+        .await?;
 
     info!("Spawning verifier update lookup table");
     let update_verifiers = verifier_lookup_table_factory
         .deploy((first_batch_size as u64, unimplemented_verifier.address()))?
         .confirmations(0usize)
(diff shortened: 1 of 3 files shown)


iamrecursion (approved):
Looks fine, but how did we let the tests get lost like this?

tests/common/chain_mock.rs
```diff
@@ -126,31 +130,24 @@ pub async fn spawn_mock_chain(
...
-    let (insert_verifiers, update_verifiers, identity_manager_impl) =
-        tokio::join!(insert_verifiers, update_verifiers, identity_manager_impl);
-
-    let insert_verifiers = insert_verifiers?;
-    let update_verifiers = update_verifiers?;
-    let identity_manager_impl = identity_manager_impl?;
+        .send()
+        .await?;
 
-    // TODO: This is sequential but could be parallelized.
```
gswirski, line 151: this TODO should stay? The new impl looks very sequential?

gswirski (approved):
nit: bring back TODO comment?

tests/common/chain_mock.rs
Dzejkop: Yes, but that was the issue. Parallelizing it caused nonce issues. It's safer to just run everything sequentially until we have a good idea on how to handle that

PR #506, opened 2023-05-09
feat: removed no longer used deployment workflow from main branch

jazzbee:
There are two reasons why we should remove it:
1. We are no longer deploying from this branch (main)
2. All deployments are moved from this repository


gswirski (approved):
😍

PR #507, opened 2023-05-09
support for dynamic batch sizes at runtime

TheMenko:
This PR is a follow up to #485

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -229,7 +229,10 @@ impl App {
         }
 
         if !self.identity_manager.has_provers().await {
-            warn!(?commitment, "Identity Manager has no provers.");
+            warn!(
+                ?commitment,
+                "Identity Manager has no provers. Add provers with /addBatchSize request."
+            );
             return Err(ServerError::NoProversOnIdInsert);
         }
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -129,12 +129,10 @@ impl App {
         let identity_manager = Arc::new(identity_manager);
         let database = Arc::new(db);
 
-        let prover_history = database.get_prover_history().await?;
+        let provers = database.get_provers().await?;
 
         // restore prover state from previus shutdown
-        identity_manager
-            .restore_prover_history(prover_history)
-            .await?;
+        identity_manager.restore_provers(provers).await?;
 
         // Await for all pending transactions
         identity_manager.await_clean_slate().await?;
--- a/src/app.rs
+++ b/src/app.rs
@@ -131,7 +131,7 @@ impl App {
 
         let provers = database.get_provers().await?;
 
-        // restore prover state from previus shutdown
+        // restore prover state from previous shutdown
         identity_manager.restore_provers(provers).await?;
 
         // Await for all pending transactions
--- a/src/app.rs
+++ b/src/app.rs
@@ -129,6 +129,13 @@ impl App {
         let identity_manager = Arc::new(identity_manager);
         let database = Arc::new(db);
 
+        let prover_history = database.get_prover_history().await?;
+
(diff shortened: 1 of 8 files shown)


src/app.rs
```diff
@@ -129,6 +129,11 @@ impl App {
         let identity_manager = Arc::new(identity_manager);
         let database = Arc::new(db);
 
+        let provers = database.get_provers().await?;
+
+        // restore prover state from previus shutdown
```
gswirski, line 134: nit: `previous`

src/database/mod.rs
```diff
@@ -325,6 +327,74 @@ impl Database {
...
+        let mut tx = self.pool.begin().await?;
+
+        let query = sqlx::query(
+            r#"
+              DELETE FROM provers WHERE batch_size = $1  
+            "#,
+        )
+        .bind(batch_size as i64);
+
+        tx.execute(query).await?;
```
gswirski, line 393: nit: do we need an explicit transaction here? wouldn't `self.pool.execute(query)` work the same?

gswirski (changes requested):
We should think about the migration plan. With this PR, there are two ways to initialize provers: from the ENV var, and from the database. There should be only one source of truth - the database.

Since we don't want to restart our production service unnecessarily (currently it takes minutes), we need to be careful how we proceed with the deployment of this. My suggestion would be:
* Keep the ENV var setting for now. When initializing the identiy_manager, we try to `restore_provers` from the db first. If and only if we get 0 provers, we fetch them from the ENV var (and insert into the db).
* Once all environments have been properly migrated (i.e. they have provers in the database), we remove the ENV var altogether
* To make future deployments and local setup easier, /insertIdentity endpoint should warn that there are no provers available and hint at addProver admin endpoint.

src/database/mod.rs
```diff
@@ -325,6 +327,69 @@ impl Database {
...
+                prover::Prover {
+                    url,
+                    batch_size,
+                    timeout_s,
+                }
+            })
+            .collect::<prover::Provers>())
+    }
+
+    pub async fn write_prover(
```
gswirski, line 356: nit: `insert_prover` or even `insert_prover_configuration` to be more consistent with other db methods

src/database/prover.rs
```diff
@@ -0,0 +1,7 @@
+pub type Provers = Vec<Prover>;
+
+pub struct Prover {
```
gswirski, line 3: nit: maybe `ProverConfiguration` for clarity?

gswirski (changes requested):
Tagging @cichaczem to confirm that it's OK to allow only one URL per batch size (given he's experimenting with more prover replicas).

I think this PR still doesn't quite work. We have a set of existing provers in the ENV var. We deploy this change, restart the process... and now we still only have those provers in the ENV var. We can't use the API to create those provers in the database because we'll end up adding duplicate batch_sizes to the identity manager.

IMHO we should do it such that if we read prover configuration from the ENV var, we migrate it to the db right away. The next time sequencer starts, the ENV is no longer used. Does it makes sense to you, @cichaczem ? I'm assuming the next PR would completely remove dependence on ENV var once all sequencer instances have been updated with this.

cichaczem:
Regarding `only one URL per batch size`. We should deploy additional instances using the infra setup if we want multiple provers for the same batch size. K8s can manage deployment and load balancing for us using 1-2 settings in the deployment description. In that manner, we should have only one URL per batch size.

cichaczem:
Ditto on @gswirski comment about populating DB with the list of provers in the ENV. Let's deploy that first, make sure we have every prover in, and remove the ENV in a separate PR.

gswirski (approved):

PR #514, opened 2023-05-16
Separate identity commitment from API process

cichaczem:
## Motivation
Currently, the app startup takes between 3 and 6 minutes. It causes problems for other systems that rely on the sequencer being available. We want to speed up the startup process to below 10 seconds.

This PR
- Losens the dependency on the local tree from `/insertIdentity` endpoint by removing pending proof from returned values
- Saves new identities to the new table
- Adds `New` and `Failed` statuses for identity
- Run local tree initialization in a separate thread blocking identity committer
- Change identity committer so it processes identities from the `unprocessed_identities` table

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Updated the changelog

### DIFF
--- a/tests/validate_proofs.rs
+++ b/tests/validate_proofs.rs
@@ -13,6 +13,8 @@ async fn validate_proofs() -> anyhow::Result<()> {
     let mut ref_tree = PoseidonTree::new(SUPPORTED_DEPTH + 1, ruint::Uint::ZERO);
     let initial_root: U256 = ref_tree.root().into();
 
+    let batch_timeout_seconds: u64 = 1;
+
     #[allow(clippy::cast_possible_truncation)]
     let tree_depth: u8 = SUPPORTED_DEPTH as u8;
     let batch_size = 3;
@@ -40,6 +42,10 @@ async fn validate_proofs() -> anyhow::Result<()> {
         "--prover-urls",
         &prover_mock.arg_string(),
         "--batch-timeout-seconds",
+        &format!("{batch_timeout_seconds}"),
+        "--dense-tree-prefix-depth",
+        "10",
+        "--tree-gc-threshold",
         "1",
     ])
     .expect("Failed to create options");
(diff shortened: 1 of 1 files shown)


src/app.rs
```diff
@@ -249,10 +256,8 @@ impl App {
     /// Will return `Err` if identity is already queued, or in the tree, or the
     /// queue malfunctions.
     #[instrument(level = "debug", skip_all)]
-    pub async fn insert_identity(
-        &self,
-        commitment: Hash,
-    ) -> Result<InclusionProofResponse, ServerError> {
+    pub async fn insert_identity(&self, commitment: Hash) -> Result<EmptyResponse, ServerError> {
```
Dzejkop, line 259: Wouldn't `()` work instead of `EmptyResponse`?

src/app.rs
```diff
@@ -274,7 +279,14 @@ impl App {
...
 
-        let (tx, rx) = oneshot::channel();
+        let identity_exists = self.database.identity_exists(commitment).await?;
+        if identity_exists {
+            return Err(ServerError::DuplicateCommitment);
+        }
+
+        self.database.insert_new_identity(commitment).await?;
+
+        let (tx, _) = oneshot::channel();
```
Dzejkop, line 289: We should get rid of this channel entirely and not send anything from the insert_identities task

schemas/database/003_unprocessed_identities.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE unprocessed_identities (
+    commitment    BYTEA       NOT NULL UNIQUE,
+    status        VARCHAR(50) NOT NULL,
+    created_at    TIMESTAMPTZ NOT NULL,
+    processed_at  TIMESTAMPTZ,
+    error_message TEXT
```
Dzejkop, line 6: Do we ever assign this error message?

src/app.rs
```diff
@@ -274,7 +279,14 @@ impl App {
...
         self.insert_identities_sender
```
Dzejkop, line 290: Also, there's now no need for this channel either. Since we put identities into the db. The `insert_identities` task can just grab them from there

Dzejkop (changes requested):
Left some comments

schemas/database/003_unprocessed_identities.sql
cichaczem: Now we do
https://github.com/worldcoin/signup-sequencer/pull/514/files#diff-23b4c110a535fa6df354073bab85cf1645400d08a9a202311db1d2898dc406bbR509
https://github.com/worldcoin/signup-sequencer/pull/514/files#diff-7f8025fc93b75296009eb2b41e0f99750760bf93b5115e03362e909036490109R70

Dzejkop (approved):

Dzejkop (approved):

PR #515, opened 2023-05-17
Removing sensitive info from logs

TheMenko:
This PR introduces SecretUrl that prevents log macro with exposing username:password

### DIFF
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -36,7 +36,7 @@ impl FromStr for SecretUrl {
     type Err = <Url as FromStr>::Err;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(Url::from_str(s).map(SecretUrl::new)?)
+        Url::from_str(s).map(SecretUrl::new)
     }
 }
 
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -119,9 +119,8 @@ mod tests {
 
     #[test]
     fn test_url_expose() {
-        let secret = SecretUrl::new(Secret(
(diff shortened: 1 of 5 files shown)


src/secret.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.format())
+    }
+}
+
+impl fmt::Debug for SecretUrl {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.format())
+    }
+}
```
Dzejkop, line 53: Not that I'm a big fan of the debug format of the `Url` type but I feel like it'd be best if we could preserve the different behaviours of debug and display for SecretUrl

Could we make it so that `.format()` returns `Url` with the username and password redacted but still use the `Url` impls of `Debug` and `Display`?

src/secret.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct SecretUrl(Url);
+
+impl SecretUrl {
+    pub fn new(url: Url) -> Self {
+        Self(url)
+    }
+
+    pub fn expose(&self) -> &str {
```
Dzejkop, line 12: Why not just return `&Url`?

Dzejkop: That's fair

cichaczem (approved):

Dzejkop (approved):
LGTM

PR #540, opened 2023-06-08
Add tree restore

TheMenko:
This PR depends on [this semaphore PR](https://github.com/worldcoin/semaphore-rs/pull/45).
The merge should be done after semaphore PR is merged.

Adds functions to attempt and restore a tree, if it fails it builds a new one

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -157,14 +157,17 @@ impl App {
             database.mark_root_as_mined(&root_hash).await?;
         }
 
+        let mmap_file_path: String = String::from("./testfile");
+
         let timer = Instant::now();
-        let tree_state = Self::initialize_tree(
+        let tree_state = Self::restore_or_initialize_tree(
             &database,
             // Poseidon tree depth is one more than the contract's tree depth
             identity_manager.tree_depth(),
             options.dense_tree_prefix_depth,
             options.tree_gc_threshold,
             identity_manager.initial_leaf_value(),
+            mmap_file_path,
         )
         .await?;
         info!("Tree state initialization took: {:?}", timer.elapsed());
@@ -199,12 +202,80 @@ impl App {
         Ok(app)
     }
 
+    async fn restore_or_initialize_tree(
+        database: &Database,
+        tree_depth: usize,
+        dense_prefix_depth: usize,
+        gc_threshold: usize,
+        initial_leaf_value: Hash,
+        mmap_file_path: String,
+    ) -> AnyhowResult<TreeState> {
+        let tree_state = match Self::get_cached_tree_state(
+            database,
+            tree_depth,
+            dense_prefix_depth,
+            gc_threshold,
+            &initial_leaf_value,
+            &mmap_file_path,
+        )
+        .await?
+        {
+            Some(state) => state,
+            None => match Self::initialize_tree(
+                database,
+                tree_depth,
+                dense_prefix_depth,
+                gc_threshold,
+                initial_leaf_value,
+                mmap_file_path,
+            )
+            .await
+            {
+                anyhow::Result::Ok(state) => state,
+                Err(e) => return Err(e),
+            },
+        };
+
+        Ok(tree_state)
+    }
+
+    #[allow(unused)]
+    async fn get_cached_tree_state(
+        database: &Database,
+        tree_depth: usize,
+        dense_prefix_depth: usize,
+        gc_threshold: usize,
+        initial_leaf_value: &Hash,
+        mmap_file_path: &String,
+    ) -> anyhow::Result<Option<TreeState>> {
+        let mined_builder = match CanonicalTreeBuilder::restore(
+            tree_depth,
+            dense_prefix_depth,
+            initial_leaf_value,
+            gc_threshold,
+            mmap_file_path,
+        ) {
+            Some(tree) => tree,
+            None => return Ok(None),
+        };
+
+        let (mined, batching_builder) = mined_builder.seal();
+        let (batching, mut latest_builder) = batching_builder.seal_and_continue();
+        let pending_items = database.get_commitments_by_status(Status::Pending).await?;
+        for update in pending_items {
+            latest_builder.update(&update);
+        }
+        let latest = latest_builder.seal();
+        Ok(Some(TreeState::new(mined, batching, latest)))
+    }
+
     async fn initialize_tree(
         database: &Database,
         tree_depth: usize,
         dense_prefix_depth: usize,
         gc_threshold: usize,
         initial_leaf_value: Hash,
+        mmap_file_path: String,
     ) -> AnyhowResult<TreeState> {
         let mut mined_items = database.get_commitments_by_status(Status::Mined).await?;
         let initial_leaves = if mined_items.is_empty() {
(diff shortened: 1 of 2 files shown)


cichaczem (commented):
The only question I have regarding mmap and restoring state came to me when I talked with the infra team about adding disk storage to sequencers. They told about instances where disks had broken data on them. It can happen because AWS is running its process like defragmentation, which can cause problems.
The question is: How will we know if the data we read is correct?

src/app.rs
```diff
@@ -199,12 +204,93 @@ impl App {
...
+        gc_threshold: usize,
+        initial_leaf_value: &Hash,
+        mmap_file_path: &String,
+    ) -> anyhow::Result<Option<TreeState>> {
+        let mut mined_items = database.get_commitments_by_status(Status::Mined).await?;
+        let last_mined = if mined_items.is_empty() {
+            initial_leaf_value.clone()
+        } else {
+            mined_items.sort_by_key(|item| item.leaf_index);
+            mined_items.last().map(|item| item.element).unwrap()
```
cichaczem, line 261: The entire `else` can be replaced with one DB query `get_latest_element`. `There's no need to fetch all mined identities to get one element.

cichaczem (approved):
LGTM

One item of work you can consider are tests for tree restoration

src/app.rs
TheMenko: I'm not sure if you referred to existing function since i couldn't locate it, so i wrote one. Could you review again now?

cichaczem (approved):

src/app.rs
```diff
@@ -101,6 +101,9 @@ pub struct Options {
     #[clap(long, env, default_value = "20")]
     pub dense_tree_prefix_depth: usize,
 
+    #[clap(long, env, default_value = "./dense_tree_mmap")]
```
Dzejkop, line 104: Could you add a doc string here? They show up in cli --help output

src/app.rs
```diff
@@ -199,12 +203,90 @@ impl App {
...
+                    info!("tree initialization successful");
+                    state
+                }
+                Err(e) => {
+                    return Err(e);
+                }
+            },
+        };
+
+        Ok(tree_state)
```
Dzejkop, line 248:
```suggestion
        if let Some(state) = Self::get_cached_tree_state(
            database,
            tree_depth,
            dense_prefix_depth,
            gc_threshold,
            &initial_leaf_value,
            &mmap_file_path,
        )
        .await?
        {
            info!("tree restored from cache");

            return Ok(AnyhowResult);
            state
        }

        let tree_state = Self::initialize_tree(
            database,
            tree_depth,
            dense_prefix_depth,
            gc_threshold,
            initial_leaf_value,
            mmap_file_path,
        )
        .await?;

        info!("tree initialization successful");

        Ok(tree_state)
```

Dzejkop, line 248:
```suggestion
        if let Some(tree_state) = Self::get_cached_tree_state(
            database,
            tree_depth,
            dense_prefix_depth,
            gc_threshold,
            &initial_leaf_value,
            &mmap_file_path,
        )
        .await?
        {
            info!("tree restored from cache");

            return Ok(tree_state);
        }

        let tree_state = Self::initialize_tree(
            database,
            tree_depth,
            dense_prefix_depth,
            gc_threshold,
            initial_leaf_value,
            mmap_file_path,
        )
        .await?;

        info!("tree initialization successful");

        Ok(tree_state)
```

src/app.rs
```diff
@@ -199,12 +203,90 @@ impl App {
...
+        dense_prefix_depth: usize,
+        gc_threshold: usize,
+        initial_leaf_value: &Hash,
+        mmap_file_path: &str,
+    ) -> anyhow::Result<Option<TreeState>> {
+        let last_mined = match database.get_last_commitment_by_status(Status::Mined).await {
+            Some(lm) => lm,
+            None => (0, *initial_leaf_value),
+        };
+
```
Dzejkop, line 263:
```suggestion
        let last_mined =
            if let Some(lm) = database.get_last_commitment_by_status(Status::Mined).await {
                lm
            } else {
                (0, *initial_leaf_value)
            };

```

src/database/mod.rs
```diff
@@ -288,6 +288,23 @@ impl Database {
...
     }
 
+    pub async fn get_last_commitment_by_status(&self, status: Status) -> Option<(usize, Hash)> {
+        let query = sqlx::query(
+            r#"
+                SELECT leaf_index, commitment
+                FROM identities
+                WHERE status = $1
+                ORDER BY leaf_index DESC
+                LIMIT 1
```
Dzejkop, line 298:
```suggestion
                LIMIT 1
```

Dzejkop, line 298:
```suggestion
                SELECT MAX(leaf_index)
                FROM identities
                WHERE status = $1
```

?

Didn't test it yet as I can't get the correct rust version

tests/dynamic_batch_sizes.rs
```diff
@@ -53,6 +53,8 @@ async fn dynamic_batch_sizes() -> anyhow::Result<()> {
         "10",
         "--tree-gc-threshold",
         "1",
+        "--dense-tree-mmap-file",
+        "./testfile",
```
Dzejkop, line 57: This will breat acceptance tests - they're treated as separate binaries and run concurrently. Can you get a temporary file path using https://crates.io/crates/tempfile ?

TheMenko: Wouldn't `#[serial_test::serial]` prevent that? That crate seems cool, i'll see what i can do

Dzejkop: I think you'd have to use https://docs.rs/serial_test/latest/serial_test/attr.file_serial.html, but also I'd like these tests to run in parallel - they quite long so we could be looking at a difference between 30-45 seconds to run the whole test stuie to like 1m 30s.

btw. this is technically undefined behaviour - since multiple processes will try to mutably map the same file to memory

src/database/mod.rs
Dzejkop: Nevermind, that only gets index

tests/dynamic_batch_sizes.rs
TheMenko: Couldn't use temp file, but instead used temp directory. It creates a dir in /tmp and then testfile is made inside. 
On drop it will delete the directory, so i think its fine solution. Also all directories are differently named (random gen).

src/app.rs
```diff
@@ -199,18 +204,106 @@ impl App {
...
+        mmap_file_path: &str,
+    ) -> anyhow::Result<Option<TreeState>> {
+        let mut last_index_in_dense: usize = 0;
+        let leftover_items = if mined_items.is_empty() {
+            vec![]
+        } else {
+            let max_leaf = mined_items.last().map(|item| item.leaf_index).unwrap();
+            // if the last index is greater then dense_prefix_depth, 1 << dense_prefix_depth
+            // should be the last index in restored tree
+            last_index_in_dense = std::cmp::min(max_leaf, 1 << dense_prefix_depth);
```
Dzejkop, line 262: Shouldn't it be `(1 << dense_prefix_depth) - 1`?

Dzejkop: Oh ok, I see. The code is correct I think, but the name is confusing. Can you rename it to something like `first_index_outside_dense`?

Dzejkop: Do you think it's feasible to write a test case for this? I'm fairly confident that this is correct but I'd be more comfortable with a test

TheMenko: that's not correct naming as well since its not always outside of dense tree. It could be first_free_index or something like that?

Dzejkop (approved):

PR #541, opened 2023-06-12
Panic on register identities tx error

cichaczem:
## Motivation

We've seen errors caused by OZ returning `out of gas` errors on identity registration. When that happened, the sequencer tried to send the next batch, even though the previous one errored. That's because we don't have a retry at this level. To mitigate this problem, we'll panic when this problem happens. This will ensure we retry the faulty batch.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/task_monitor/tasks/process_identities.rs
+++ b/src/task_monitor/tasks/process_identities.rs
@@ -364,14 +364,15 @@ async fn commit_identities(
         )
         .await;
 
-    if let Err(err) = transaction_result {
-        panic!(
-            "Failed to insert identity to contract due to error {err}. Restarting sequencer to \
-             reconstruct local tree"
-        );
-    }
-
-    let transaction_id = transaction_result.unwrap();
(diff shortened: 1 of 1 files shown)


src/task_monitor/tasks/process_identities.rs
```diff
@@ -354,19 +354,27 @@ async fn commit_identities(
...
+
+    if transaction_result.is_err() {
+        transaction_result.map_err(|e| {
             error!(?e, "Failed to insert identity to contract.");
             e
         })?;
+        panic!(
+            "Failed to insert identity to contract. Restarting sequencer to reconstruct local tree"
+        );
+    }
```
Dzejkop, line 375:
```suggestion
    if let Some(err) = transaction_result {
        panic!(
            "Failed to insert identity to contract due to error {err}. Restarting sequencer to reconstruct local tree"
        );
    }
```

Dzejkop (approved):

PR #554, opened 2023-07-11
Make sequencer multi-chain : Adds "finalized" status

Dzejkop:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


kustosz (approved):
Looks good, but I'd be careful when deploying this for the first time – you'll end up loading whole tree into sparse, then moving it to dense element by element

PR #562, opened 2023-07-17
Build and run inside docker image

Dzejkop:
## Motivation
Proposal rendered as image:
Untitled Diagram drawio

Essentially, right now our build pipeline is pretty complex:
1. We're doing a lot of work to be able to build `arm64` images, but then we never end up using them.
2. We're building static binaries, even though we're still running them inside containers (kind defeats the purpose)
3. We have 2 separate Dockerfiles, 3 places in which we set the Rust version.
4. We have to setup QEMU on our pipeline in order to build for the proper architecture - again, never using the problematic architecture in the end.

What I'm proposing is a drastic simplification:
A single Dockerfile in the sequencer repo
A single place to set the Rust version
No explicit support for arm64
We build the binary inside the container in which it will be run
We build the binary "the normal way", i.e. with dynamic linking

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/schemas/openapi.yaml
+++ b/schemas/openapi.yaml
@@ -82,6 +82,13 @@ paths:
           application/json:
             schema:
               $ref: "#/components/schemas/VerifySemaphoreProofRequest"
+      parameters:
+        - in: query
+          name: maxRootAgeSeconds
+          schema:
+            type: integer
+          description: The max age in seconds of the provided root.
+                       If the root is older than this value this endpoint will return an error and a 400 response.
       responses:
         '200':
           description: Valid proof response
--- a/src/app.rs
+++ b/src/app.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use anyhow::Result as AnyhowResult;
+use chrono::Duration;
 use clap::Parser;
 use hyper::StatusCode;
 use semaphore::poseidon_tree::LazyPoseidonTree;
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -1,52 +1,38 @@
...
+COPY . .
+
+# Install dependencies
+RUN apt-get update && \
+    apt-get install -y curl build-essential libssl-dev texinfo libcap2-bin pkg-config
 
-ARG BIN=rust-app
+# TODO: Use a specific version of rustup
+# Install rustup
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
```
Dzejkop, line 16: I'm installing rustup here instead of using a rust base container, because I want to be able to only set the versino in `rust-toolchain.toml`

cichaczem (approved):
Let's try it!

Dockerfile
```diff
@@ -1,52 +1,38 @@
...
-STOPSIGNAL SIGTERM
-HEALTHCHECK NONE
-ENTRYPOINT ["/bin"]
+RUN /src/target/release/signup-sequencer --version
+
+# cc variant because we need libgcc and others
+FROM gcr.io/distroless/cc-debian11:nonroot
+
+# Copy the sequencer binary
+COPY --from=build-env --chown=0:10001 --chmod=001 /src/target/release/signup-sequencer /bin/signup-sequencer
```
jazzbee, line 36:
```suggestion
COPY --from=build-env --chown=0:10001 --chmod=010 /src/target/release/signup-sequencer /bin/signup-sequencer
```

jazzbee: Execute permission must be set for a group, not the others.

PR #574, opened 2023-07-19
remove env provers

TheMenko:


tests/multi_prover.rs
```diff
@@ -83,6 +83,22 @@ async fn multi_prover() -> anyhow::Result<()> {
...
+        batch_timeout_seconds as usize,
+    )
+    .await?;
+    test_add_prover(
+        &uri,
+        &client,
+        &prover_map[&batch_size_10],
+        batch_timeout_seconds as usize,
+    )
+    .await?;
```
Dzejkop, line 100: Could you create a function that accepts a list (mapping?) of provers and adds all of them to given sequencer?

Actually since we're mocking provers, we could probably extract their creation from `spawn_deps` and move it after the sequencer is up and running

tests/common/mod.rs
```diff
@@ -336,6 +331,26 @@ fn construct_verify_proof_body(
     )
 }
 
+fn construct_add_batch_size_body(prover: &ProverService, timeout_seconds: usize) -> Body {
+    Body::from(
+        json!({
```
Dzejkop, line 336: `AddBatchSizeRequest` is public in signup-sequencer, we can use that instead

PR #583, opened 2023-08-28
Added/updated database tables to support deletion/recovery.

0xKitsune:
This PR adds a `deletions` and `recoveries` table which can be seen below.

```sql
CREATE TABLE deletions (
    leaf_index    BIGINT NOT NULL PRIMARY KEY
)

CREATE TABLE recoveries (
    existing_commitment    BYTEA       NOT NULL UNIQUE,
    new_commitment         BYTEA       NOT NULL UNIQUE
)
```

The `identities` table was updated to include an `id` which is now the primary key that auto increments upon each addition into the table.

```sql
-- Add the new 'id' column
ALTER TABLE identities ADD COLUMN id BIGSERIAL;

-- Set the new 'id' column as NOT NULL
ALTER TABLE identities ALTER COLUMN id SET NOT NULL;

-- Set the id to be unique
ALTER TABLE identities ADD CONSTRAINT id_unique UNIQUE(id);

-- Drop the existing primary key
ALTER TABLE identities DROP CONSTRAINT identities_pkey;

-- Set the new 'id' column as the primary key
ALTER TABLE identities ADD PRIMARY KEY (id);
```

Lastly, an `eligibility` timestamp was added to the `unprocessed_identities` table. This allows the sequencer to know when to insert a new identity commitment during the recovery process, avoiding users having multiple valid commitments at one time.


schemas/database/004_identities.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Add the new 'id' column
+ALTER TABLE identities ADD COLUMN id BIGSERIAL;
```
Dzejkop, line 2: We should also make sure that the initial value of `id` is the same as `leaf_index`

src/database/mod.rs
```diff
@@ -492,6 +493,73 @@ impl Database {
...
+
+        let result = self.pool.fetch_all(query).await?;
+
+        Ok(result
+            .into_iter()
+            .map(|row| row.get::<i64, _>(0))
+            .collect::<Vec<i64>>())
+    }
+
     pub async fn get_unprocessed_commitments(
```
Dzejkop, line 563: You probably want to modify this method to only grab elligible identities

src/database/mod.rs
```diff
@@ -642,12 +712,13 @@ mod test {
     async fn setup_db() -> anyhow::Result<(Database, DockerContainerGuard)> {
         let db_container = postgres_docker_utils::setup().await?;
         let port = db_container.port();
+        dbg!(&db_container.port());
```
Dzejkop, line 715: leftover dbg

schemas/database/006_deletions.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE deletions (
+    leaf_index    BIGINT NOT NULL PRIMARY KEY
```
Dzejkop, line 2: You might consider adding a commitment column here. It'll make it easier to correlate the replacement identity

Dzejkop: Especially given the fact that `leaf_index` won't be unique any longer

Dzejkop (approved):

PR #584, opened 2023-08-29
Add option to validate semaphore proof with root age

Dzejkop:
## PR Checklist

- [x] Added Tests
- [x] Added Documentation

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -16,7 +16,7 @@ use crate::database::prover::{ProverConfiguration as DbProverConf, Provers};
 use crate::database::{self, Database};
 use crate::ethereum::{self, Ethereum};
 use crate::identity_tree::{
-    CanonicalTreeBuilder, Hash, InclusionProof, RootItem, Status, TreeState,
+    CanonicalTreeBuilder, Hash, InclusionProof, RootItem, Status, TreeState, TreeVersionReadOps,
 };
 use crate::prover::batch_insertion::ProverConfiguration;
 use crate::prover::map::make_insertion_map;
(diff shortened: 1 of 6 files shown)


dcbuild3r (commented):
lgtm

src/app.rs
```diff
@@ -461,6 +468,52 @@ impl App {
...
+            }
+            _ => (),
+        }
+
+        // Latest processed root is always valid
+        if root_state.root == processed_root {
+            return Ok(());
+        }
+
+        let max_root_age = Duration::seconds(max_root_age_seconds);
```
0xKitsune, line 496: Can this be passed in as `Duration`?

src/app.rs
```diff
@@ -461,6 +468,52 @@ impl App {
...
+
+        let mined_at = root_state.mined_valid_as_of;
+        let pending_as_of = root_state.pending_valid_as_of;
+
+        let now = chrono::Utc::now();
+
```
0xKitsune, line 502: Do you need to initialize varibles here? Or can you just simplify to the following:

```rust
let root_age = if root_state.status == Status::Pending {
    chrono::Utc::now() - root_state.pending_valid_as_of;
} else {
    let mined_at = root_state.mined_valid_as_of.ok_or(ServerError::InvalidRoot)?;
    chrono::Utc::now() - mined_at
};
```

src/app.rs
0xKitsune, line 496: Can this be passed in as `Duration`?

src/app.rs
Dzejkop: Fair, but I'll leave the `now` variable.

0xKitsune (approved):

PR #586, opened 2023-08-30
Feat: Deletion/Recovery API Updates

0xKitsune:
This PR adds support for deletion provers, enabling provers to be added and removed via the `addBatch` and `removeBatch` endpoints. Additionally, this PR separates deletion provers from insertion provers within the `IdentityManager`, allowing for a prover instance to specify separate batch sizes for deletions and insertions.


src/prover/mod.rs
```diff
@@ -7,9 +7,631 @@
 //! APIs are designed to be imported for use qualified (e.g.
 //! `batch_insertion::Prover`, `batch_insertion::Identity` and so on).
 
-pub mod batch_insertion;
+pub mod identity;
 pub mod map;
 pub mod proof;
 
+use std::collections::HashSet;
```
Dzejkop, line 14: I'm assuming not much changed in this file from it's previous version under `batch_insertion/mod.rs`, right?

tests/common/prover_mock.rs
```diff
@@ -83,18 +83,37 @@ impl ProveResponse {
...
+//TODO: we could just import this from the sequencer
+#[derive(Debug, Copy, Clone, sqlx::Type, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+#[sqlx(type_name = "prover_enum", rename_all = "PascalCase")]
+pub enum ProverType {
+    #[default]
+    Insertion,
+    Deletion,
+}
+impl std::fmt::Display for ProverType {
```
Dzejkop, line 110:
```suggestion
}

impl std::fmt::Display for ProverType {
```

src/app.rs
```diff
@@ -317,23 +324,103 @@ impl App {
...
+        // TODO: update this to check if there are deletion provers
+        if !self.identity_manager.has_provers().await {
+            warn!(
+                ?commitment,
+                "Identity Manager has no provers. Add provers with /addBatchSize request."
+            );
+            return Err(ServerError::NoProversOnIdInsert);
+        }
+
+        let identity_exists = self.database.identity_exists(commitment).await?;
```
Dzejkop, line 354: Something to keep in mind is that this function simply checks for the existance of the given commitment in the identities table.

But with the new schema an identity could have been already deleted.

src/app.rs
```diff
@@ -133,10 +132,10 @@ pub struct Options {
 }
 
 pub struct App {
-    database:           Arc<Database>,
-    identity_manager:   SharedIdentityManager,
+    database: Arc<Database>,
```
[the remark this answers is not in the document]
0xKitsune: Looks like I was using a different version and the repo is using
```
  RUST_VERSION: "1.65"
  NIGHTLY_VERSION: nightly-2022-11-04
```

I updated my env and ran `cargo fmt --all`.

src/prover/mod.rs
0xKitsune: Yes, the functionality did not change. I moved the logic within `batch_insertion/mod.rs` into `prover/mod.rs`.

schemas/database/008_prover_type.sql
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+
+
+
+
+
+
+
```
Dzejkop, line 28: why the newlines?

0xKitsune: This was unintentional, removing now.

src/app.rs
```diff
@@ -317,23 +326,109 @@ impl App {
...
+        }
+
+        // Ensure that deletion provers exist
+        if !self.identity_manager.has_deletion_provers().await {
+            warn!(
+                ?new_commitment,
+                "Identity Manager has no deletion provers. Add provers with /addBatchSize request."
+            );
+            return Err(ServerError::NoProversOnIdDeletion);
+        }
```
Dzejkop, line 411: That's redundant because of `self.delete_identity` no?

0xKitsune: Yes, good catch. Removing this snippet now.

src/identity_tree.rs
```diff
@@ -183,6 +183,11 @@ where
         self.tree.root()
     }
 
+    /// Gets the leaf value at a given index.
+    fn get_leaf(&self, leaf: usize) -> Hash {
+        self.get_leaf(leaf)
```
Dzejkop, line 188:
```
self.tree.get_leaf(leaf)
```

otherwise you'll have infinite recursion

PR #589, opened 2023-09-04
Deletion/recovery task monitor updates

0xKitsune:
This PR introduces the necessary updates to `TaskMonitor` to enable deletion/recovery features.


src/contracts/mod.rs
```diff
@@ -174,13 +174,27 @@ impl IdentityManager {
...
+    }
+
     #[instrument(level = "debug", skip(prover, identity_commitments))]
     pub async fn prepare_insertion_proof(
         prover: ReadOnlyInsertionProver<'_>,
         start_index: usize,
         pre_root: U256,
-        post_root: U256,
         identity_commitments: &[Identity],
+        post_root: U256,
```
Dzejkop, line 197: Nice!

src/contracts/mod.rs
```diff
@@ -204,6 +218,26 @@ impl IdentityManager {
         Ok(proof_data)
     }
 
+    #[instrument(level = "debug", skip(prover, identity_commitments))]
+    pub async fn prepare_deletion_proof(
```
Dzejkop, line 222: The signature for getting the proof will be different. Instead of commitments we'll use leaf indexes.

Refer to https://github.com/worldcoin/semaphore-mtb/blob/87c50e4eb609e6fa80e70ff16e8d00b93deeb384/prover/deletion_proving_system.go#L16

src/database/mod.rs
```diff
@@ -578,12 +578,31 @@ impl Database {
...
             .map(|row| DeletionEntry {
-                leaf_index: row.get::<i64, _>(0),
+                leaf_index: row.get::<i64, _>(0) as usize,
                 commitment: row.get::<Hash, _>(1),
             })
             .collect::<Vec<DeletionEntry>>())
     }
 
+    /// Remove a list of entries from the deletions table
+    pub async fn remove_deletions(&self, commitments: Vec<Hash>) -> Result<(), Error> {
```
Dzejkop, line 588: Could you write a test for this function?

I have a feeling you'd need `where commitment in $1` if you're using `push_values`

src/prover/mod.rs
```diff
@@ -331,6 +405,16 @@ struct ProofInput {
     merkle_proofs:        Vec<Vec<U256>>,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct DeletionProofInput {
+    input_hash:           U256,
+    pre_root:             U256,
+    post_root:            U256,
+    identity_commitments: Vec<U256>,
```
Dzejkop, line 414: You'll also need indexes here

src/task_monitor/tasks/delete_identities.rs
```diff
@@ -0,0 +1,146 @@
...
+// be compliant
+async fn delete_identities(
+    database: &Database,
+    identity_manager: &IdentityManager,
+    latest_tree: &TreeVersion<Latest>,
+    deletion_time_interval: u64,
+    min_deletion_batch_size: usize,
+    wake_up_notify: Arc<Notify>,
+) -> AnyhowResult<()> {
+    // TODO: do we need to do this for this step?
```
Dzejkop, line 74: No, we don't :)

src/task_monitor/tasks/delete_identities.rs
```diff
@@ -0,0 +1,146 @@
...
+
+            // Ensure that the next leaf matches in the db and the latest tree
+            let next_db_index = database.get_next_leaf_index().await?;
+            let next_leaf = latest_tree.next_leaf();
+
+            assert_eq!(
+                next_leaf, next_db_index,
+                "Database and tree are out of sync. Next leaf index in tree is: {next_leaf}, in \
+                 database: {next_db_index}"
+            );
```
Dzejkop, line 113: We don't need this check here

src/contracts/mod.rs
Dzejkop: Correction: we'll need both the commitments and indexes

src/task_monitor/tasks/delete_identities.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::contracts::{IdentityManager, SharedIdentityManager};
+use crate::database::types::DeletionEntry;
+use crate::database::Database;
+use crate::identity_tree::{
+    Canonical, Hash, Intermediate, Latest, TreeVersion, TreeVersionReadOps, TreeWithNextVersion,
+};
+use crate::task_monitor::PendingBatchSubmission;
+use crate::utils::async_queue::{AsyncPopGuard, AsyncQueue};
+
+// It can take up to 40 minutes to bridge the root
```
Dzejkop, line 20: Leftover

src/task_monitor.rs
```diff
@@ -239,6 +254,24 @@ impl TaskMonitor {
...
         handles.push(insert_identities_handle);
 
+        // // Delete identities task
+        let delete_identites = DeleteIdentities::new(
+            self.database.clone(),
+            self.identity_manager.clone(),
+            self.tree_state.get_latest_tree(),
+            self.deletion_time_interval,
+            self.min_batch_deletion_size,
+            wake_up_notify,
```
Dzejkop, line 264: This `Notify` gets triggered when something is inserted via the api. Can you make sure it's also triggered on deletion or add a new notify just for deletion?

src/identity_tree.rs
```diff
@@ -194,15 +194,22 @@ where
...
+        let first_is_zero = match next.metadata.diff.first() {
+            Some(first) => first.update.element == Hash::ZERO,
+            None => return vec![],
+        };
+
+        // Gets the next contiguous of insertion or deletion updates from the diff
         next.metadata
             .diff
             .iter()
+            .take_while(|&update| (update.update.element == Hash::ZERO) == first_is_zero)
```
Dzejkop, line 212: Could you write a test for this function?

Dzejkop: Also how about this:
```rust
let should_take = |elem: &&AppliedTreeUpdate| {
    if first_is_zero {
        // If first is zero, we should take only consecutive zeros
        elem.update.element == Hash::ZERO
    } else {
        // If first is not zero, we should take only non-zeros
        elem.update.element != Hash::ZERO
    }
};
```

and then

```rust
.take_while(should_take)
```

?

src/task_monitor/tasks/insert_identities.rs
```diff
@@ -42,7 +42,9 @@ async fn insert_identities_loop(
 ) -> AnyhowResult<()> {
     loop {
         // get commits from database
-        let unprocessed = database.get_unprocessed_commitments(Status::New).await?;
+        let unprocessed = database
+            .get_eligible_unprocessed_commitments(Status::New)
```
Dzejkop, line 46: We can now remove `get_unprocessed_commitments`

src/task_monitor/tasks/process_identities.rs
```diff
@@ -13,17 +15,20 @@ use tracing::{debug, error, info, instrument, warn};
...
 use crate::prover::map::ReadOnlyInsertionProver;
+use crate::prover::{Proof, Prover, ReadOnlyProver};
 use crate::task_monitor::{PendingBatchSubmission, TaskMonitor};
 use crate::utils::async_queue::AsyncQueue;
 
 /// The number of seconds either side of the timer tick to treat as enough to
 /// trigger a forced batch insertion.
 const DEBOUNCE_THRESHOLD_SECS: u64 = 1;
 
+const RECOVERY_WAIT_TIME: Days = Days::new(7);
```
Dzejkop, line 30: I think ideally we should read this from chain. We could even read it on every batch I think - we're producing them pretty infrequently.

src/task_monitor/tasks/process_identities.rs
```diff
@@ -203,14 +199,73 @@ async fn process_identities(
...
     identity_manager: &IdentityManager,
     batching_tree: &TreeVersion<Intermediate>,
     pending_batch_submissions_queue: &AsyncQueue<PendingBatchSubmission>,
     updates: &[AppliedTreeUpdate],
-    insertion_prover: ReadOnlyInsertionProver<'_>,
+) -> AnyhowResult<()> {
+    // If the update is an insertion
+    if updates
+        .first()
+        .expect("TODO: propagate error, updates should be > 1")
```
Dzejkop, line 212:
```suggestion
        .context("TODO: propagate error, updates should be > 1")?
```

src/task_monitor/tasks/process_identities.rs
```diff
@@ -394,3 +445,199 @@ async fn commit_identities(
...
+            start_index,
+        })
+        .await;
+
+    // Update the batching tree only after submitting the identities to the chain
+    batching_tree.apply_updates_up_to(post_root.into());
+
+    info!(start_index, ?pre_root, ?post_root, "Tree updated");
+
+    // TODO: check if any deleted commitments correspond with entries in the
```
Dzejkop, line 615: This part should happen in `mine_identities`

If we do it here we could have an overlap of how long it takes to mine a transaction on chain when both the deleted identity and it's replacement is usable.

Not sure what we should do in case of a reorg though.

src/task_monitor/tasks/delete_identities.rs
```diff
@@ -0,0 +1,146 @@
...
+) -> AnyhowResult<()> {
+    // TODO: do we need to do this for this step?
+    info!("Awaiting for a clean slate");
+    identity_manager.await_clean_slate().await?;
+
+    info!("Starting deletion processor.");
+
+    let deletion_time_interval = Duration::from_secs(deletion_time_interval);
+
+    // TODO: we should track this with persistance to eunsure we are deleting at
```
Dzejkop, line 82: That's a good point. This timeout will probably be very large in prod deployment.

We could estimate the last batch time by fetching the latest root in root history and take it's expiry timestamp?

0xKitsune: I added a new table that only allows for one row to track the latest deletion.

Issue #590, opened 2023-09-05
Scan chain events instead of attempting to finalize each root


Dzejkop:
Currently the sequencer takes each and every root that has been mined on L1 and polls each supported L2 contract to figure out if the given root is either the latest root on that contract or it's present in root history.

With the new bridging setup we might have roots which are missing on L2 - but that shouldn't stop the sequencer from moving forward.

Breakdown:
1. [ ] Scan events on L1 and L2 to monitor roots being added, we can use [get_logs](https://docs.rs/ethers/1.0.2/ethers/providers/struct.Provider.html#method.get_logs). Note that events will be different on L1 contracts and on L2 contracts
2. [ ] Parse the logs and extract post roots
3. [ ] Once a root has been found in the logs from all supported chains - proceed as previously, i.e. mark the root as finalized/mined and update the mined tree.
4. [ ] Cleanup:
         The sequencer should keep track of the roots extracted from logs in step 2 (in memory). But once a given root is finalized on all the chains - roots which are older than this root should be removed from this in-memory storage.
         The sequencer has access to the list of all roots so if we finalize a root N+1, we can see that root N is older and can remove it from the list/mapping.

Some notes:
1. [ ] It's not clear to me how we should determine the block from which to start scanning on each chain - grabbing the latest root at sequencer startup seems reasonable but it means we might miss some roots.

PR #591, opened 2023-09-07
Dzejkop/scan-roots

Dzejkop:
This resolves https://github.com/worldcoin/signup-sequencer/issues/590

Now instead of trying to finalize each root in sequence - we parse a continuous stream of RootAdded or TreeChanged events and on each one we attempt to finalize the given root.


src/contracts/mod.rs
```diff
@@ -64,6 +65,15 @@ pub struct IdentityManager {
 }
 
 impl IdentityManager {
+    // TODO: I don't like these public getters
```
0xKitsune, line 68: Agreed, we can either make the `abi` and `secondary_abis` fields public or create specific functions to abstract logic that interacts with the abis.

src/contracts/scanner.rs
```diff
@@ -0,0 +1,58 @@
...
+{
+    pub async fn new_latest(read_provider: T, window_size: u64) -> anyhow::Result<Self> {
+        let latest_block = read_provider.get_block_number().await?;
+
+        Ok(Self {
+            read_provider,
+            current_block: latest_block.as_u64(),
+            window_size,
+        })
+    }
```
0xKitsune, line 23: If we ever want to start from an earlier block than the latest, consider changing this to 

```rust
pub async fn new_latest(read_provider: T, window_size: u64, block_number: Option<BlockNumber>) -> anyhow::Result<Self> {
    let block_number = if let Some(block_number){
        block_number
    } else{
        read_provider.get_block_number().await?
    };

    Ok(Self {
        read_provider,
        current_block: block_number.as_u64(),
        window_size,
    })
}
```

src/contracts/scanner.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    pub async fn next(
+        &mut self,
+        address: Option<ValueOrArray<Address>>,
+        topics: [Option<Topic>; 4],
+    ) -> anyhow::Result<Vec<Log>> {
+        let latest_block = self.read_provider.get_block_number().await?;
+
+        let latest_block = latest_block.as_u64();
```
0xKitsune, line 32: Can simplify into 

`let latest_block = self.read_provider.get_block_number().await?.as_u64();`

src/contracts/scanner.rs
```diff
@@ -0,0 +1,58 @@
...
+                block_option: FilterBlockOption::Range {
+                    from_block,
+                    to_block,
+                },
+                address,
+                topics,
+            })
+            .await?;
+
+        self.current_block = next_current_block;
```
0xKitsune, line 54: Can simplify into 

`self.current_block = to_block + 1;`

which allows you to remove 

`let next_current_block = to_block + 1;`

src/task_monitor/tasks/finalize_identities.rs
```diff
@@ -56,61 +57,152 @@ async fn finalize_roots_loop(
...
+        secondary_logs.extend(logs);
     }
+
+    let mut all_roots = vec![];
+
+    let mainnet_roots = extract_root_from_mainnet_logs(&mainnet_logs);
+    let secondary_roots = extract_roots_from_secondary_logs(&secondary_logs);
+
+    all_roots.extend(mainnet_roots);
+    all_roots.extend(secondary_roots);
```
0xKitsune, line 134: Can be simplified to 

```rust

    let mut roots = extract_root_from_mainnet_logs(&mainnet_logs);
    roots.extend(extract_roots_from_secondary_logs(&secondary_logs));
```

src/task_monitor/tasks/finalize_identities.rs
```diff
@@ -56,61 +57,152 @@ async fn finalize_roots_loop(
...
+    mainnet_scanner: &mut BlockScanner<A>,
+    secondary_scanners: &mut HashMap<Address, BlockScanner<B>>,
+    mainnet_address: Address,
+) -> anyhow::Result<Vec<U256>>
+where
+    A: Middleware,
+    <A as Middleware>::Error: 'static,
+    B: Middleware,
+    <B as Middleware>::Error: 'static,
+{
```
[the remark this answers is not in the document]
Dzejkop: Differnet types unfortunately - in practice one is `ReadProvider` and the other is `Arc<ReadProvider>`, which is annyoing but it is what it is

src/contracts/scanner.rs
Dzejkop: Not quite, because `to_block` is an `Option<BlockNumber>` at this time. And I wanted to not update the current block unless we succeed in fetching logs

src/contracts/mod.rs
Dzejkop: Yeah, let's leave the TODO for now. I'm not supper happy with this struct nor the `Ethereum` struct so a bigger refactor would be in order I think

src/contracts/scanner.rs
Dzejkop: If we want to start from a specific block number I think I'd want a new `new` function which accepts specific block number. Which would make this function infallible and non-const

src/contracts/scanner.rs
0xKitsune: sgtm

0xKitsune (approved):

0xKitsune:
lgtm pending test patches

src/task_monitor/tasks/finalize_identities.rs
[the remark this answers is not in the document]
Dzejkop: I suppose I could go with `T` and `Arc<T>` but 2 generics are more flexible

PR #593, opened 2023-09-08
Deletion proof padding

0xKitsune:
This PR introduces logic to pad deletion proofs. The verifier and prover can only work with a given batch size, so we need to ensure that our batches match that size. We do this by padding deletion indices with tree.depth() ^ 2. The deletion prover will skip the proof for any deletion with an index greater than the max tree depth.

### DIFF
--- a/src/task_monitor/tasks/process_identities.rs
+++ b/src/task_monitor/tasks/process_identities.rs
@@ -7,7 +7,7 @@ use chrono::{Days, Utc};
 use ethers::types::U256;
 use once_cell::sync::Lazy;
 use prometheus::{register_histogram, Histogram};
-use semaphore::poseidon_tree::{Branch, PoseidonHash};
+use semaphore::poseidon_tree::Branch;
 use tokio::sync::Notify;
 use tokio::{select, time};
 use tracing::{debug, error, info, instrument, warn};
--- a/src/task_monitor/tasks/process_identities.rs
+++ b/src/task_monitor/tasks/process_identities.rs
@@ -505,14 +505,12 @@ pub async fn delete_identities(
         .collect::<Vec<u32>>();
 
     // The verifier and prover can only work with a given batch size, so we need to
-    // ensure that our batches match that size. We do this by padding with
-    // subsequent zero identities and their associated merkle proofs if the batch is
-    // too small.
-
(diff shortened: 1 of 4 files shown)


src/task_monitor/tasks/process_identities.rs
```diff
@@ -477,7 +477,7 @@ pub async fn delete_identities(
 
     // Next get merkle proofs for each update - note the proofs are acquired from
     // intermediate versions of the tree
-    let mut merkle_proofs: Vec<_> = updates
+    let merkle_proofs: Vec<_> = updates
```
Dzejkop, line 480: Don't we need to pad out merkle proofs with essentially all zero bytes?

I know we say that prover will skip the proofs but afaik it's a kind of a branchless programming situation and you do actually this data to be there

src/task_monitor/tasks/process_identities.rs
```diff
@@ -499,42 +499,24 @@ pub async fn delete_identities(
...
-    if commitment_count != batch_size {
-        // FIXME: I dont think this is going to work for deletions
-        // TODO: need to check this, but basically we are getting the most recent leaf
-        // and then using that as the start since deletions can happen out of sequence
-
-        // TODO: Calling count is inefficient because it calls `next()` until the
-        // iterator is exhausted. Update this to be more efficient
-        let start_index = latest_tree_from_updates.leaves().count();
-        let padding = batch_size - commitment_count;
-        commitments.append(&mut vec![U256::zero(); padding]);
```
Dzejkop, line 515: As above - we also need to pad commitments

src/task_monitor/tasks/process_identities.rs
```diff
@@ -499,42 +501,34 @@ pub async fn delete_identities(
...
 
-        // TODO: need to check these indices are correct
-        for i in start_index..(start_index + padding) {
-            let proof = latest_tree_from_updates.proof(i);
-            merkle_proofs.push(proof);
-        }
-    }
+        let zeroed_proof = Proof(vec![
+            Branch::Left(Uint::ZERO);
+            latest_tree_from_updates.depth()
```
Dzejkop, line 522: log2 of this?

Dzejkop (approved):

PR #605, opened 2023-09-20
Fix: Sync tree state / `get_commitments_by_status` 

0xKitsune:
This PR introduces changes to the identities table to fix tree state syncing on startup. Additionally, `get_commitments_by_status` was updated to only get the `TreeUpdate` at a given leaf index with the highest `id` value.

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -1417,7 +1417,8 @@ mod test {
         let (db, _db_container) = setup_db().await?;
 
         let identities = mock_identities(5);
-        let roots = mock_roots(5);
+
+        let roots = mock_roots(7);
 
         for i in 0..5 {
             db.insert_pending_identity(i, &identities[i], &roots[i])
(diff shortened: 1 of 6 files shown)


tests/delete_identities.rs
```diff
@@ -145,6 +145,33 @@ async fn delete_identities() -> anyhow::Result<()> {
...
+
+    // Test loading the state from a file when the on-chain contract has the state.
+    let (app, local_addr) = spawn_app(options.clone())
+        .await
+        .expect("Failed to spawn app.");
+    let uri = "http://".to_owned() + &local_addr.to_string();
+
+    // Ensure that identities have been deleted
+    for i in 0..deletion_batch_size {
+        test_inclusion_proof(
```
Dzejkop, line 163: Can you also add inclusion proofs for identities which have not been deleted?

src/database/mod.rs
```diff
@@ -312,10 +312,10 @@ impl Database {
     ) -> Result<Vec<TreeUpdate>, Error> {
         let query = sqlx::query(
             r#"
-            SELECT leaf_index, commitment
+            SELECT DISTINCT ON (leaf_index) leaf_index, commitment
```
Dzejkop, line 315: Let's add a test that we return only the latest identities

Dzejkop (approved):

PR #611, opened 2023-09-25
tree mmap cache

TheMenko:
Rebasing older branch was too troublesome, so i made a new one

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -354,7 +354,6 @@ impl App {
                 }
             }
             None => {
-                // compare to initial root?
                 if !mined.get_root().eq(&initial_root_hash) {
                     return Ok(None);
                 }
--- a/src/app.rs
+++ b/src/app.rs
@@ -16,7 +16,8 @@ use crate::contracts::{IdentityManager, SharedIdentityManager};
 use crate::database::{self, Database};
 use crate::ethereum::{self, Ethereum};
 use crate::identity_tree::{
-    CanonicalTreeBuilder, Hash, InclusionProof, RootItem, Status, TreeState, TreeVersionReadOps,
+    CanonicalTreeBuilder, Hash, InclusionProof, RootItem, Status, TreeState, TreeUpdate,
+    TreeVersionReadOps,
 };
 use crate::prover::map::initialize_prover_maps;
 use crate::prover::{self, ProverConfiguration, ProverType, Provers};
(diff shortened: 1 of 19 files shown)


src/app.rs
```diff
@@ -235,15 +241,133 @@ impl App {
...
+        mmap_file_path: &str,
+    ) -> anyhow::Result<Option<TreeState>> {
+        let mut last_mined_index_in_dense: usize = 0;
+        let leftover_items = Self::get_leftover_leaves_and_update_index(
+            &mut last_mined_index_in_dense,
+            dense_prefix_depth,
+            mined_items,
+        );
 
+        let Some(mined_builder) = CanonicalTreeBuilder::restore(
```
Dzejkop, line 330: You need this line:
```
        let mined_items = dedup_tree_updates(mined_items);
```

somewhere here, otherwise the `mined_items` can contain duplicate items

src/identity_tree.rs
```diff
@@ -673,6 +675,50 @@ impl CanonicalTreeBuilder {
...
+                dense_prefix_depth,
+                initial_leaf,
+                mmap_file_path,
+            ) {
+                Ok(tree) => tree,
+                Err(error) => {
+                    warn!("Tree wasn't restored. Reason: {}", error.to_string());
+                    return None;
+                }
+            };
```
Dzejkop, line 699: What happens if the tree restores from the file but the values are actually different than what we expect?

I don't think this can currently happen, because at startup we only mark as "processed"not "mined", but technically such a situation would be possible if the sequencer crashed after submitting a deletion batch and then restarted with the updated values (if we ever start marking as "mined" at startup).

src/app.rs
```diff
@@ -235,15 +241,133 @@ impl App {
...
+            initial_leaf_value,
+            last_mined_index_in_dense,
+            &leftover_items,
+            gc_threshold,
+            mmap_file_path,
+        ) else {
+            return Ok(None);
+        };
+
+        let (mined, mut processed_builder) = mined_builder.seal();
```
Dzejkop, line 342: Related to [this comment](https://github.com/worldcoin/signup-sequencer/pull/611/files#r1341289604) I think we should add a check above this line that the latest expected root from the db matches the root given to us by this restored tree

src/database/mod.rs
```diff
@@ -331,6 +333,19 @@ impl Database {
             .collect::<Vec<_>>())
     }
 
+    pub async fn get_latest_root_by_status(&self, status: Status) -> Result<Option<Hash>, Error> {
+        let query = sqlx::query(
+            r#"
+              SELECT root from identities where status = $1 ORDER BY leaf_index DESC LIMIT 1
```
Dzejkop, line 339:
```suggestion
            SELECT root 
            FROM identities
            WHERE status = $1
            ORDER BY id DESC
            LIMIT 1
```

Fixed capitalization and changed `leaf_index` to `id` to the order by. We've introduced id as part of deletion/recovery as now we can have updates essentially at any leaf in the tree.

Can you double check that this works?

PR #627, opened 2023-10-04
add insertion timestamp persistence

dcbuild3r:
## Motivation

If the sequencer halts, the in-memory representation of the last time a batch was inserted is lost, therefore we need to introduce `latest_insertion_timestamp` persistence by writing it to the database.

## Solution

Creates a new table called `latest_insertion_timestamp` and writes the last time a batch was inserted into the merkle tree into `insertion_timestamp`. The value is read from `process_identities` on each call and writes to it after a batch has been committed. 

## PR Checklist

- [x] Added Tests

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -369,8 +369,9 @@ impl Database {
     pub async fn get_latest_insertion_timestamp(&self) -> Result<Option<DateTime<Utc>>, Error> {
         let query = sqlx::query(
             r#"
-            SELECT insertion_timestamp FROM latest_insertion_timestamp
-            "#,
+            SELECT insertion_timestamp 
+            FROM latest_insertion_timestamp 
+            WHERE Lock = 'X';"#,
         );
 
         let row = self.pool.fetch_optional(query).await?;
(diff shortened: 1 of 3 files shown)


src/database/mod.rs
```diff
@@ -537,6 +549,22 @@ impl Database {
...
 
+    pub async fn update_latest_insertion_timestamp(
+        &self,
+        insertion_timestamp: DateTime<Utc>,
+    ) -> Result<(), Error> {
+        let query = sqlx::query(
+            r#"
+            INSERT INTO latest_insertion_timestamp (insertion_timestamp)
+            VALUES ($1)
+            "#,
```
0xKitsune, line 560: This will add a new entry into the table every time this function is called instead of updating the existing entry. 

To update the existing entry, you could an "id" field and make sure to update that specific id each time. Alternatively, you could add a constraint similar to the latest deletion table.

dcbuild3r: noted, I constrained it similarly like with the deletion root

Dzejkop (approved):
LGTM

PR #633, opened 2023-10-13
Strengthen postgres-docker-utils

Dzejkop:
Changes the naive logic to `SocketAddr` parsing - we'll grab and use the first available socket address exposed by docker


dcbuild3r (approved):
LGTM

PR #635, opened 2023-10-13
convert leaf_index to ruint u256

smatthewenglish:
initial attempt at addressing the following TODO in the file `src/database/mod.rs`

```
// TODO: consider using a larger value than i64 for leaf index, ruint should
// have postgres compatibility for u256
```

please let me know if this is useful and I'll be happy to keep making progress on it

the approach I've taken is to store `leaf_index` as a `BYTEA` in the database, because a `BIGINT` would be too small

there are [some parts of the code](https://github.com/worldcoin/signup-sequencer/compare/main...smatthewenglish:u256-leaf-index#diff-0b979b6de560b29c1d97336878db56179224aa21d96fe099630c6628479ed020R258) that use the value of `leaf_index` to set the size of/index into an array, which could create problems if the `U256` value of `leaf_index` is bigger than `usize`

very appreciative of any and all feedback

bullish on worldcoin! would love to help out this initiative 🙌

### DIFF
--- a/schemas/database/001_init.sql
+++ b/schemas/database/001_init.sql
@@ -1,5 +1,5 @@
 CREATE TABLE identities (
-    leaf_index    BIGINT      NOT NULL PRIMARY KEY,
+    leaf_index    BYTEA       NOT NULL PRIMARY KEY,
     commitment    BYTEA       NOT NULL UNIQUE,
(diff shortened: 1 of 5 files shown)


Dzejkop:
Hey, thanks for the PR but the comment is actually outdated. `i64` will likely cover our needs many times over

PR #638, opened 2023-10-23
Audit-changes

Dzejkop:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/identity_tree/status.rs
+++ b/src/identity_tree/status.rs
@@ -9,11 +9,11 @@ use thiserror::Error;
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
 #[serde(rename_all = "camelCase")]
 pub enum Status {
-    /// The root is included in sequencer's in-memory tree, but is not yet
+    /// Root is included in sequencer's in-memory tree, but is not yet
     /// mined.
     Pending,
 
-    /// The root is mined on mainnet but is still waiting for confirmation on
+    /// Root is mined on mainnet but is still waiting for confirmation on
     /// relayed chains
     ///
     /// i.e. the root is included in a mined block on mainnet,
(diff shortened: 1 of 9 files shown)


dcbuild3r (approved):
LGTM

src/identity_tree/status.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// status should immediately become Finalized
+    Processed,
+
+    /// Root is mined and relayed to secondary chains
+    Mined,
+}
+
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "camelCase")]
+pub enum UnprocessedStatus {
```
0xKitsune, line 32: Consider adding a doc comment for this struct to provide clarity on the difference between `Status` and `UnprocessedStatus`.

src/identity_tree/status.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// A status type visible on the API level - contains both the processed and
+/// unprocessed statuses
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[serde(rename_all = "camelCase")]
+#[serde(untagged)]
+pub enum ApiStatus {
+    Unprocessed(UnprocessedStatus),
+    Processed(Status),
+}
```
0xKitsune, line 51: If this pertains to the API/server, what do you think about moving it into the server module rather than in the identity tree?

Dzejkop: It's used by the api but not only - it's just a way to unify these 2 types but also keep them separate where it's important

Dzejkop: I suppose we could make it so that
`ApiStatus -> Status`
`Status -> ProcessedStatus` but I'm not sure if I'm a huge fan of that

Dzejkop: Actually I changed it to just that

0xKitsune (approved):

PR #651, opened 2023-11-15
Dzejkop/identity-history

Dzejkop:
A feature for deletion/recovery

The purpose is to provide a detailed history view into a given identity. Format of the history response is the following
```json
{
  "history": [
    { "kind": "insertion", "status": "buffered" }
  ]
}
```

Where:
1. There can be at most 2 entries in the history (1 for insertion, 1 for deletion)
2. Kind can be `insertion` or `deletion`
3. Status can be `bufferd`, `queued`, `pending`, `batched`, `mined`, `bridged` - for more info look into `IdentityHistoryEntryStatus` in `src/app.rs`

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -9,7 +9,7 @@ use hyper::StatusCode;
 use ruint::Uint;
 use semaphore::poseidon_tree::LazyPoseidonTree;
 use semaphore::protocol::verify_proof;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use tracing::{info, instrument, warn};
 
 use crate::contracts::{IdentityManager, SharedIdentityManager};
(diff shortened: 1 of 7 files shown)


src/app.rs
```diff
@@ -637,6 +564,59 @@ impl App {
...
+                }
+                _ => (),
+            }
+
+            let kind = if entry.commitment == Uint::ZERO {
+                IdentityHistoryEntryKind::Deletion
+            } else {
+                IdentityHistoryEntryKind::Insertion
+            };
+
```
0xKitsune, line 613: You could move this logic into an Into<T> implementation, not necessary though.

0xKitsune: Sgtm, disregard.

0xKitsune (approved):

PR #662, opened 2023-12-14
Remove `UnprocessedStatus::Failed`

0xKitsune:
This PR removes the `Failed` variant from the `UnprocessedStatus` enum. 

While this PR does not remove the `UnprocessedStatus` enum entirely, it seems that `Unprocessed` could be combined with the `ProcessedStatus` variants to have one cohesive `Status` enum ([link for reference](https://github.com/worldcoin/signup-sequencer/blob/main/src/identity_tree/status.rs#L10-L52)).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -443,6 +443,8 @@ impl App {
             return Err(ServerError::UnreducedCommitment);
         }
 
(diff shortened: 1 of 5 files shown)


Dzejkop (approved):
LGTM

PR #674, opened 2024-01-17
Refactor: Config & simplified tasks

Dzejkop:
Includes:
1. Deduplication in process_identities
2. Replaces CLI options with `config.rs` - now sequencer can be configured with env vars or a config file (or both) - NOTE: env vars will have changed names now
3. Simplification of tasks (removed a lot of redundant code)
4. Replaced the parsed config in tests with a type safe config builder
5. Removed some unused files
6. Removed a lot of type aliases


src/secret.rs
```diff
@@ -1,9 +1,11 @@
 use std::fmt;
 use std::str::FromStr;
 
+use serde::{Deserialize, Serialize};
 use url::Url;
 
-#[derive(Clone, Eq, PartialEq)]
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(transparent)]
```
0xKitsune, line 8: Consider moving `secret.rs` to utils.

src/serde_utils.rs
```diff
@@ -1,11 +1,11 @@
+use std::borrow::Cow;
```
0xKitsune, line 1: Consider moving this to `utils` as well.

src/server.rs
```diff
@@ -1,21 +1,19 @@
 pub mod error;
 
-use std::net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener};
+use std::net::TcpListener;
```
0xKitsune, line 3: Consider moving this to `server` and renaming the file to `mod.rs`.

src/server.rs
```diff
@@ -148,32 +133,11 @@ async fn list_batch_sizes(
...
-    };
-    let port = options.server.port().unwrap_or(9998);
-    let addr = SocketAddr::new(ip, port);
-
-    info!("Will listen on {}", addr);
-    let listener = TcpListener::bind(addr)?;
-
-    let serve_timeout = Duration::from_secs(options.serve_timeout);
-    bind_from_listener(app, serve_timeout, listener).await?;
+pub async fn main(app: Arc<App>, config: ServerConfig) -> anyhow::Result<()> {
```
0xKitsune, line 136: Consider renaming to `run` or similar to avoid having more than one `main`.

src/main.rs
```diff
@@ -2,15 +2,61 @@
...
+    task_monitor.start().await;
+
+    // Start server (will stop on shutdown signal)
+    server::main(app_for_server, server_config).await?;
+
+    tracing::info!("Stopping the app");
+    task_monitor.shutdown().await?;
+
+    Ok(())
+}
```
0xKitsune, line 58: Consider using only `eyre::Result` instead of both `eyre` and `anyhow::Result`. I think we can refactor to something like the following:

```rust


async fn app(args: Args) -> eyre::Result<()> {
    let mut settings = config::Config::builder();

    if let Some(path) = args.config {
        settings = settings.add_source(config::File::from(path).required(true));
    }

    let settings = settings
        .add_source(config::Environment::with_prefix("SEQ").separator("__"))
        .build()?;

    let config = settings.try_deserialize::<Config>()?;

    let server_config = config.server.clone();

    // Create App struct
    let app = Arc::new(App::new(config).await?);
    let app_for_server = app.clone();

    let task_monitor = TaskMonitor::new(app);

    // Process to push new identities to Ethereum
    task_monitor.start().await;

    // Start server (will stop on shutdown signal)
    server::main(app_for_server, server_config).await?;

    tracing::info!("Stopping the app");
    task_monitor.shutdown().await?;

    Ok(())
}

```

src/task_monitor.rs
```diff
@@ -126,55 +72,15 @@ pub struct TaskMonitor {
...
-    scanning_window_size:           u64,
-    scanning_chain_head_offset:     u64,
-    time_between_scans:             Duration,
-    max_epoch_duration:             Duration,
-    // TODO: docs
-    batch_deletion_timeout_seconds: i64,
-    // TODO: docs
-    min_batch_deletion_size:        usize,
-    monitored_txs_capacity:         usize,
+    instance: RwLock<Option<RunningInstance>>,
```
0xKitsune, line 75: Consider moving `handles` and `shutdown_sender` directly into `TaskMonitor`, removing `RunningInstance`.

src/utils.rs
```diff
@@ -7,7 +7,9 @@ use tokio::sync::broadcast;
 use tokio::task::JoinHandle;
 use tracing::{error, info};
 
+pub mod batch_type;
```
0xKitsune, line 10: Consider renaming this file and moving into `utils`.

src/prover/batch_insertion/mod.rs
```diff
@@ -1,588 +0,0 @@
...
-    /// # Arguments
-    /// - `start_index`: The index in the merkle tree at which the insertions
-    ///   were started.
-    /// - `pre_root`: The value of the merkle tree's root before identities were
-    ///   inserted.
-    /// - `post_root`: The value of the merkle tree's root after the identities
-    ///   were inserted.
-    /// - `identities`: A list of identity insertions, ordered in the order the
-    ///   identities were inserted into the merkle tree.
-    pub async fn generate_proof(
```
0xKitsune, line 143: Do we need this functionality for tests? It looks like its not used anywhere in the core codebase.

src/prover/batch_insertion/mod.rs
```diff
@@ -1,588 +0,0 @@
...
-///   insertions started.
-/// - `PreRoot` is `pre_root`, the root value of the merkle tree before the
-///   insertions were made.
-/// - `PostRoot` is `post_root`, the root value of the merkle tree after the
-///   insertions were made.
-/// - `IdComms` is `identity_commitments`, the list of identity commitments
-///   provided in the order that they were inserted into the tree.
-///
-/// The result is computed using the inputs in _big-endian_ byte ordering.
-pub fn compute_input_hash(
```
0xKitsune, line 225: Same here, do we need this for tests? If not, then we should be fine to remove.

0xKitsune (approved):
LGTM, left some minor comments.

src/main.rs
Dzejkop: Yes, one thing at a time though - I'll also want to remvoe cli-batteries and use telemetry-batteries instead.

src/server.rs
Dzejkop: I'm actually more partial to self named mod files and I'm trying to progressively move away from mod.rs

src/utils.rs
Dzejkop: This is already in utils

src/server.rs
ewoolsey: > I'm actually more partial to self named mod files and I'm trying to progressively move away from mod.rs

Oh yeah? Why do you prefer that? I sorta always liked the ‘mod.rs’ method for how it contains a module inside a directory.

src/task_monitor.rs
Dzejkop: Eventually I want to get rid of TaskMonitor entirely - or modify it significantly. But one thing at a time

src/prover/map.rs
```diff
@@ -1,69 +1,52 @@
...
     /// Removes the prover for the provided `batch_size` from the prover map.
-    pub fn remove(&mut self, batch_size: usize) -> Option<P> {
-        self.map.remove(&batch_size)
+    pub fn remove(&mut self, batch_size: usize) -> Option<Prover> {
+        self.map.remove(batch_size)
     }
 
     pub fn len(&self) -> usize {
         self.map.len()
     }
```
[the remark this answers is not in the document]
Dzejkop: That'll make all the invocations more complicated.

I would be more partial to an `AsRef` impl but that also has downsides. Let's keep is the way it is until we figure out a good abstraction.

0xKitsune (approved):

PR #709, opened 2024-04-16
Use testcontainers to run postgres database. Disable shutdown tests due to interfering tests (global var impl).

piohei:
## Motivation

As discussed:
There was a bug (docker containers running after tests) in docker wrapper and to follow direction from other projects testcontainers was introduced here instead. Docker image of postgres is being used for testing. Question: which version of postgres is being used? I think it should match the version running in prod env.
There was also a bug in shutdown testing. Tests were interfering due to shutdown being called twice (once from one test and second time from the other). It was causing panic, which was causing process abort (due to rust configuration in Cargo.toml), which was causing strange/hard to understand test results (random tests being red without any special error). This test was disabled for now.

## Solution

Introduced testcontainers. Disabled shutdown test.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/database/mod.rs
+++ b/src/database/mod.rs
@@ -860,9 +860,10 @@ mod test {
     use anyhow::Context;
     use chrono::{Days, Utc};
     use ethers::types::U256;
-    use postgres_docker_utils::DockerContainerGuard;
+    use postgres_docker_utils::DockerContainer;
     use ruint::Uint;
     use semaphore::Field;
+    use testcontainers::clients::Cli;
 
     use super::Database;
     use crate::config::DatabaseConfig;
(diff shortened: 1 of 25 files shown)


src/shutdown.rs
```diff
@@ -81,6 +81,10 @@ mod tests {
 
     use super::*;
 
+    // This test is ignored due to global variable being used to indicate if system is shutting
+    // down. Because tests are being run in parallel it is causing unpredictable behaviour and
+    // random test failures.
+    #[ignore]
```
Dzejkop, line 87: Let's just delete it

src/database/mod.rs
```diff
@@ -891,21 +892,32 @@ mod test {
...
+        DBSetuper {
+            docker: Cli::default(),
+        }
+    }
 
-        Ok((db, db_container))
+    impl DBSetuper {
+        // TODO: we should probably consolidate all tests that propagate errors to
+        // TODO: either use anyhow or eyre
+        async fn setup_db<'a>(&'a self) -> anyhow::Result<(Database, DockerContainer<'a>)> {
```
Dzejkop, line 908: Can we ditch `DBSetuper` and make it into 
```
async fn setup_db() -> anyhow::Result<(Database, DockerContainer<'a>)> {
    let docker = Cli::default();
    // ...
}
```

Dzejkop (approved):

src/shutdown.rs
piohei: Done

src/database/mod.rs
piohei: As discussed - refactored.

PR #711, opened 2024-04-22
fix(deletions): Deletions hotfix

0xKitsune:
This PR applies the hotfixes contained in [1.0.3](https://github.com/worldcoin/signup-sequencer/releases/tag/1.0.3) - [1.0.5](https://github.com/worldcoin/signup-sequencer/releases/tag/1.0.5), updating the following logic:

- Introduces `pending_insertions_mutex` to ensure that deletion and insertion updates are not interleaved into the `pending_identities` table.

- Updates `process_identities` task to process deletions immediately and process partial insertion batches when the next batch is deletion.


Dzejkop (approved):
LGTM

PR #716, opened 2024-04-26
Fix restoring tree when there are no commitments added.

piohei:
## Motivation

When empty tree was being restored from file starting index was 1 instead of 0. It shouldn't be a big deal for production but may be problematic in some test cases. And still it is a bug - so nice to be fixed.

## Solution

When tree is being restored instead of using zero as default value None optional is used to indicate that actually there is no latest index as there are no used indexes at all.

## PR Checklist

- [ X] Added Tests
- [ ] Added Documentation


tests/tree_restore_multiple_comittments.rs
```diff
@@ -0,0 +1,93 @@
...
+use signup_sequencer::identity_tree::TreeVersionReadOps;
+
+use crate::common::test_same_tree_states;
+
+mod common;
+
+const IDLE_TIME: u64 = 7;
+
+#[tokio::test]
+async fn tree_restore_multiple_commitments() -> anyhow::Result<()> {
```
Dzejkop, line 11: Typo in the file name should be `commitments`

tests/tree_restore_one_comittment.rs
```diff
@@ -0,0 +1,91 @@
...
+use signup_sequencer::identity_tree::TreeVersionReadOps;
+
+use crate::common::test_same_tree_states;
+
+mod common;
+
+const IDLE_TIME: u64 = 7;
+
+#[tokio::test]
+async fn tree_restore_one_commitment() -> anyhow::Result<()> {
```
Dzejkop, line 11: Typo in the file name, should be `commitment`

tests/tree_restore_multiple_comittments.rs
```diff
@@ -0,0 +1,93 @@
+use common::prelude::*;
+use signup_sequencer::identity_tree::TreeVersionReadOps;
+
+use crate::common::test_same_tree_states;
```
Dzejkop, line 4: Let's put the extra imports in `common::prelude`

Dzejkop (approved):

PR #718, opened 2024-04-29
High Availability - sync over DB, multiple process instances support

piohei:
## Motivation

It is required to run signup-sequencer with HA.

## Solution

Sync workers over DB.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


schemas/database/013_batches_and_transactions.sql
```diff
@@ -0,0 +1,25 @@
+-- Create ENUM for prover type
+CREATE TYPE batch_type_enum AS ENUM ('Insertion', 'Deletion');
+
+CREATE TABLE batches
+(
+    next_root    BYTEA           NOT NULL UNIQUE PRIMARY KEY,
+    prev_root    BYTEA UNIQUE,
+    created_at   TIMESTAMPTZ     NOT NULL,
```
Dzejkop, line 8: can be

```
created_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
```

that way you can omit it in inserts

schemas/database/013_batches_and_transactions.sql
```diff
@@ -0,0 +1,25 @@
+-- Create ENUM for prover type
+CREATE TYPE batch_type_enum AS ENUM ('Insertion', 'Deletion');
```
Dzejkop, line 2: Postgres enums are tricky as you can't add a variant to them later down the line.

I would suggest changing it to a number or string.

Another option is a table with allowed values but that's a bit overkill imo.

src/database/mod.rs
```diff
@@ -840,6 +833,135 @@ pub trait DatabaseExt<'a>: Executor<'a, Database = Postgres> {
...
+        next_root: &Hash,
+        batch_type: BatchType,
+        commitments: &Commitments,
+        leaf_indexes: &LeafIndexes,
+    ) -> Result<(), Error> {
+        let query = sqlx::query(
+            r#"
+            INSERT INTO batches(
+                next_root,
+                prev_root,
```
Dzejkop, line 848: we can omit prev_root here, no?

Dzejkop: I'm also fine with explicit null

schemas/database/013_batches_and_transactions.sql
```diff
@@ -0,0 +1,25 @@
...
+    FOREIGN KEY (prev_root) REFERENCES batches (next_root)
+);
+
+CREATE UNIQUE INDEX i_single_null_prev_root ON batches ((batches.prev_root IS NULL)) WHERE batches.prev_root IS NULL;
+
+CREATE TABLE transactions
+(
+    transaction_id  VARCHAR(256) NOT NULL UNIQUE PRIMARY KEY,
+    batch_next_root BYTEA        NOT NULL UNIQUE,
+    created_at      TIMESTAMPTZ  NOT NULL,
```
Dzejkop, line 22: Same as in `batches` table

src/database/mod.rs
```diff
@@ -840,6 +833,135 @@ pub trait DatabaseExt<'a>: Executor<'a, Database = Postgres> {
...
+        .bind(leaf_indexes);
+
+        self.execute(query).await?;
+        Ok(())
+    }
+
+    async fn get_next_batch(self, prev_root: &Hash) -> Result<Option<BatchEntry>, Error> {
+        let res = sqlx::query_as::<_, BatchEntry>(
+            r#"
+            SELECT * FROM batches WHERE prev_root = $1
```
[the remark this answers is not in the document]
Dzejkop: Also applies to other usages of `*`

schemas/database/013_batches_and_transactions.sql
piohei: I have seen we were using it for "prover_type" (file 008_prover_type.sql). What we prefer? Do it like for prover or just use String field and parse on rust side?

src/database/mod.rs
piohei: I personally prefer to be more explicit (not to check schema on table for every insert) but I'm fine with both approaches.

schemas/database/013_batches_and_transactions.sql
Dzejkop: Ok, good point. Let's keep it consistent

src/database/mod.rs
Dzejkop: Ok, let's leave it like it is

PR #724, opened 2024-05-07
Added batching using DB. Additional tree added to have proper state on startup.

piohei:
This is a first step to introduce batches and transactions table. This is not HA solution yet - just the first step towards it.

## Motivation

Make small changes instead of one big.

## Solution

Now processing batches is changed into 2 steps. First one is to create a batch (select identities, leaf indexes, etc.). Then the batch is saved in database. Second step is reading data from database (batch) and executes it. Proper batch execution includes storing transaction id in database.

As an alternative solution we could save in database a batch with a proves (no need to calculate it later). It will allow second step (transaction creation/execution) to not use any tree at all. It will require more memory used in database (it can be cleared from time to time as we don't need all the batches forever) but will make HA solution much easier.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


src/identity_tree.rs
```diff
@@ -263,6 +263,9 @@ impl BasicTreeOps for TreeVersionData<lazy_merkle_tree::Derived> {
     fn update(&mut self, leaf_index: usize, element: Hash) {
         let updated_tree = self.tree.update(leaf_index, &element);
 
+        if self.tree.root() == updated_tree.root() {
```
piohei, line 266: I added it here because of recreating tree on startup. It is hard to easily select from database proper groups of commitments. There are three options:
1. As done change the code to not apply updates that don't change the tree at all. Why? Because then we are peeking changes to create a batch that are not changing anything and we can't put such a batch in database (unique/fk constraints).
2. Instead of doing it here we can peek the "no-changing" changes and just not save them in database.
3. Add additional internal status just for this "batched but not processed yet" situation.
4. Do more queries on startup but may use much more memory as we are loading almost all commitments. :/

src/identity_tree.rs
```diff
@@ -280,7 +283,25 @@ impl BasicTreeOps for TreeVersionData<lazy_merkle_tree::Derived> {
     fn apply_diffs(&mut self, mut diffs: Vec<AppliedTreeUpdate>) {
         let last = diffs.last().cloned();
 
-        self.metadata.diff.append(&mut diffs);
+        if !diffs.is_empty() {
```
piohei, line 286: Same as above.

src/database/mod.rs
```diff
@@ -840,6 +880,198 @@ pub trait DatabaseExt<'a>: Executor<'a, Database = Postgres> {
...
+    }
+
+    async fn get_next_batch_without_transaction(self) -> Result<Option<BatchEntry>, Error> {
+        let res = sqlx::query_as::<_, BatchEntry>(
+            r#"
+            SELECT
+                batches.id as id,
+                batches.next_root as next_root,
+                batches.prev_root as prev_root,
+                batches.created_at as created_at,
```
Dzejkop, line 972: AFAIK sqlx handles fully qualified names, so you can leave this (and others) as `batches.created_at`

piohei: I'm sure I have done it due to created_at conflict but will reassure myself. :)

src/database/mod.rs
```diff
@@ -310,6 +311,45 @@ impl Database {
...
+
+        let batch_entry = tx.get_next_batch(last_processed_hash).await?;
+        let Some(batch_entry) = batch_entry else {
+            return Ok(vec![]);
+        };
+
+        let batches = tx.get_all_batches_after(batch_entry.id).await?;
+        let mut result = vec![];
+        for batch in batches {
+            match batch.batch_type {
```
Dzejkop, line 329: Can't we have the same code for both cases? In case for deletion we replace `*batch.commitments.0.get(i).unwrap()` with `Hash::ZERO` but `*batch.commitments.0.get(i).unwrap()` should be zero anyway, no?

Dzejkop (approved):

PR #727, opened 2024-05-10
E2E testing with docker compose

piohei:
## Motivation

This E2E testing will be used mostly for HA solution testing. We need to run multiple instances of sequencer to ensure our prod env will work properly.

## Solution

E2E tests are basicly running docker compose under the hood and execute some requests to follow different kind of scenarios. This docker compose can be also used for local development.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,135 @@
...
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=sequencer
+    ports:
+      - "5461:5432"
+    volumes:
+      - sequencer_db_data:/var/lib/postgresql/data
+  tx-sitter:
+    container_name: tx-sitter
+    image: tx-sitter-monolith
```
Dzejkop, line 36: I made the package public, we can use
`image: ghcr.io/worldcoin/tx-sitter-monolith:latest`

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,135 @@
+version: "3"
+services:
+  chain:
+    container_name: chain
+    image: ghcr.io/foundry-rs/foundry
+    hostname: chain
+    platform: linux/amd64
+    ports:
+      - "8545:8545"
```
Dzejkop, line 9: Would be nice if we could run all of these tests and envs in parallel. One thing that we'd need for that is exposing different (random) ports per test

e2e_tests/scenarios/tests/common/docker_compose.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Starts a docker compose infrastructure. It will be stopped and removed when
+/// the guard is dropped.
+///
+/// Note that we're using sync code here so we'll block the executor - but this
+/// is fine, because the spawned container will still run in the background.
+pub async fn setup(cwd: &str) -> anyhow::Result<DockerComposeGuard> {
+    let res = DockerComposeGuard { cwd };
+
+    let (stdout, stderr) = run_cmd_to_output(res.cwd, "docker compose --profile e2e-ha up -d")
```
Dzejkop, line 60: In order to run multiple envs in parallel we would want to run with `--project-name {random_generated_id_or_string}`

piohei:
I've added option for parallel run. There are few things left:
* Is it safe to commit into this repo keys for insertion prover and deletion prover?
* I'm not sure if I've done right things with committing docker compose with anvil args where I use url for fork that includes API key for alchemy.
* When do we want to run these tests? They will be the slowest ones.

Dzejkop:
> I've added option for parallel run. There are few things left:
> 
> * Is it safe to commit into this repo keys for insertion prover and deletion prover?
> * I'm not sure if I've done right things with committing docker compose with anvil args where I use url for fork that includes API key for alchemy.
> * When do we want to run these tests? They will be the slowest ones.

We can include keys no problem. The API keys is a bit problematic though. Maybe we can pass it through GH secrets?

Regarding when to run these - how about we set a cronjob and run it a couple times a day?

Dzejkop:
> how about we set a cronjob and run it a couple times a day?
alternatively everytime we push a tag?

piohei:
> > how about we set a cronjob and run it a couple times a day?
> > alternatively everytime we push a tag?

Tag sounds good to me and makes sense. We will treat it as "last checkpoint" before going on with deploy. :)

piohei:
> > I've added option for parallel run. There are few things left:
> > 
> > * Is it safe to commit into this repo keys for insertion prover and deletion prover?
> > * I'm not sure if I've done right things with committing docker compose with anvil args where I use url for fork that includes API key for alchemy.
> > * When do we want to run these tests? They will be the slowest ones.
> 
> We can include keys no problem. The API keys is a bit problematic though. Maybe we can pass it through GH secrets?
> 
> Regarding when to run these - how about we set a cronjob and run it a couple times a day?

I will set this API key as env. Then we can use GH secrets for CI. :)

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      - chain
+      - semaphore-insertion
+      - semaphore-deletion
+      - tx-sitter
+    restart: always
+    ports:
+      - ${SIGNUP_SEQUENCER_0_PORT:-9080}:8080
+    environment:
+      - RUST_LOG=debug
+      - SEQ__TREE__TREE_DEPTH=30
```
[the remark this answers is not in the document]
piohei: At first there was single file with everything configured here but now there is balancer so it makes sense to move config to file.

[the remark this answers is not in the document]
piohei: Done. ;)

piohei:
Docker compose execution downloads MTB keys if they are not present.

e2e_tests/docker-compose/keys/.keep
Dzejkop, line 1: Do we need this file? I thought we would download the keys everytime

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,147 @@
...
+      - TX_SITTER__SERVICE__PREDEFINED__NETWORK__HTTP_RPC=http://chain:8545
+      - TX_SITTER__SERVICE__PREDEFINED__NETWORK__WS_RPC=ws://chain:8545
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__ID=1b908a34-5dc1-4d2d-a146-5eb46e975830
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__NAME=Relayer
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__CHAIN_ID=31337
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__KEY_ID=d10607662a85424f02a33fb1e6d095bd0ac7154396ff09762e41f82ff2233aaa
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__API_KEY=G5CKNF3BTS2hRl60bpdYMNPqXvXsP-QZd2lrtmgctsnllwU9D3Z4D8gOt04M0QNH
+      - TX_SITTER__SERVER__HOST=0.0.0.0:3000
+      - TX_SITTER__SERVER__DISABLE_AUTH=true
+  semaphore-keys-init-insertion:
```
Dzejkop, line 56: Maybe `restart: on-failure`?

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,147 @@
...
+    image: curlimages/curl:latest
+    user: "0:0"
+    volumes:
+      - ./keys:/keys
+    entrypoint: /bin/sh
+    command: [ "-c", "if [ ! -f \"/keys/insertion_b10t30.ps\" ]; then curl \"https://semaphore-mtb-trusted-setup-ceremony.s3.amazonaws.com/insertion_b10/insertion_b10t30.ps\" -o /keys/insertion_b10t30.ps && chown -f $(stat -c \"%u:%g\" /keys/.keep) /keys/insertion_b10t30.ps; fi" ]
```
Dzejkop, line 62: Maybe

```suggestion
  semaphore-keys-init-insertion:
    image: curlimages/curl:latest
    user: "0:0"
    restart: on-failure
    volumes:
      - ./keys:/keys
    entrypoint: /bin/sh
    command: [ "-c", "if [ ! -f \"/keys/insertion_b10t30.ps\" ]; then curl \"https://semaphore-mtb-trusted-setup-ceremony.s3.amazonaws.com/insertion_b10/insertion_b10t30.ps\" -o /keys/insertion_b10t30.ps && chown -f $(stat -c \"%u:%g\" /keys/.keep) /keys/insertion_b10t30.ps; fi" ]
```

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,147 @@
...
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__KEY_ID=d10607662a85424f02a33fb1e6d095bd0ac7154396ff09762e41f82ff2233aaa
+      - TX_SITTER__SERVICE__PREDEFINED__RELAYER__API_KEY=G5CKNF3BTS2hRl60bpdYMNPqXvXsP-QZd2lrtmgctsnllwU9D3Z4D8gOt04M0QNH
+      - TX_SITTER__SERVER__HOST=0.0.0.0:3000
+      - TX_SITTER__SERVER__DISABLE_AUTH=true
+  semaphore-keys-init-insertion:
+    image: curlimages/curl:latest
+    user: "0:0"
+    volumes:
+      - ./keys:/keys
+    entrypoint: /bin/sh
```
Dzejkop, line 61: maybe?
```suggestion
    restart: on-failure
    entrypoint: /bin/sh
```

e2e_tests/docker-compose/compose.yml
Dzejkop, line 62:
```suggestion
    command:
      - "-c"
      - >
        if [ ! -f "/keys/insertion_b10t30.ps" ]; then
          curl "https://semaphore-mtb-trusted-setup-ceremony.s3.amazonaws.com/insertion_b10/insertion_b10t30.ps" -o /keys/insertion_b10t30.ps &&
          chown -f $(stat -c "%u:%g" /keys/.keep) /keys/insertion_b10t30.ps;
        fi
```

Dzejkop: Let's do the same for deletion keys

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,147 @@
...
+        condition: service_completed_successfully
+  semaphore-deletion:
+    image: semaphore-mtb
+    hostname: semaphore-deletion
+    restart: always
+    ports:
+      - ${SEMAPHORE_DELETION_PORT:-3002}:3001
+    command: [ "start", "--keys-file", "/mtb/keys", "--prover-address", "0.0.0.0:3001", "--mode", "deletion" ]
+    volumes:
+      - ./keys/deletion_b10t30.ps:/mtb/keys
```
Dzejkop, line 92:
```suggestion
      - ./keys:/mtb/keys
```

e2e_tests/docker-compose/compose.yml
piohei: Thanks!

e2e_tests/docker-compose/compose.yml
```diff
@@ -0,0 +1,147 @@
...
+    ports:
+      - ${SIGNUP_SEQUENCER_BALANCER_PORT:-8080}:8080
+    volumes:
+      - ./haproxy:/usr/local/etc/haproxy
+    depends_on:
+      - signup-sequencer-0
+  signup-sequencer-0: &signup-sequencer-def
+    image: signup-sequencer
+    hostname: signup-sequencer-0
+    profiles: [ e2e-ha ]
```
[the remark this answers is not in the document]
piohei: I added that profile so the docker compose could be used to setup local env. But now it is in e2e_tests directory and I have no idea if anyone will use it at all. What do you think? Should we add `e2e-ha` profile for balancer service also or just remove profile at all?

e2e_tests/docker-compose/keys/.keep
piohei: It is created due to 2 things:
* This is a hack to change ownership of downloaded files from root:root to same user:group as for that file (which should be normal user).
* I didn't add checking if folder exists or not in bash so this way I can be sure it is there.

e2e_tests/docker-compose/compose.yml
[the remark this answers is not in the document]
piohei: I will keep the profile to allow use of docker compose for local development also. :)

Dzejkop (approved):

e2e_tests/scenarios/tests/common/api.rs
```diff
@@ -0,0 +1,141 @@
...
+    uri: &String,
+    commitment: &String,
+) -> anyhow::Result<()> {
+    debug!("Calling /insertIdentity");
+    let body = Body::from(
+        json!({
+            "identityCommitment": format!("0x{}", commitment),
+        })
+        .to_string(),
+    );
```
0xKitsune, line 27: Lets use `InsertCommitmentRequest` and serialize into a string.

e2e_tests/scenarios/tests/common/api.rs
```diff
@@ -0,0 +1,141 @@
...
+    client: &Client<HttpConnector>,
+    uri: &String,
+    commitment: &String,
+) -> anyhow::Result<()> {
+    debug!("Calling /deleteIdentity");
+    let body = Body::from(
+        json!({
+            "identityCommitment": format!("0x{}", commitment),
+        })
+        .to_string(),
```
0xKitsune, line 65: Lets use `DeletionRequest` and serialize into string.

e2e_tests/scenarios/tests/common/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+    for _ in 0..count {
+        // Generate the identities using the just the last 64 bits (of 256) has so we're
+        // guaranteed to be less than SNARK_SCALAR_FIELD.
+        let commitment = Hash::from(rand::random::<u64>());
+
+        commitments.push(commitment);
+    }
+
+    commitments
+}
```
0xKitsune, line 99: Nit: Can be simplified to 

```rust

PR #736, opened 2024-05-24
Retry transactions that fail due to serializable isolation level

ewoolsey:
Simple macro that help build and retry transactions that fail during tx.commit(). The first commit is just splitting up database into multiple files so it's likely better to view each individual commit to find the important changes.


Dzejkop (approved):
LGTM

PR #738, opened 2024-05-27
Added retrieving identities in batches.

piohei:
## Motivation

Retrieving few millions rows in single query may not be a great idea. This change introduces retrieving them in batches (20k each batch).

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation


piohei:
Sqlx is doing it under the hood. No need for custom impl.

PR #742, opened 2024-06-04
Introduce transaction manager

piohei:
## Motivation

It is hard to introduce a new storage for transactions (off chain mode) with current state of code. Disabling some tasks for off-chain mode is not enough to make it work (initialization phase, connection to relayer, etc).

## Solution

This PR is about extracting transactions layer to separate types for future work where we will allow different storage for transactions (not only chain as now). There are also small changes to unify how tasks are written in terms of waiting between executions.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


src/task_monitor/tasks/delete_identities.rs
```diff
@@ -21,53 +23,60 @@ pub async fn delete_identities(
...
-
-            // Insert the new items into pending identities
-            let items = data.into_iter().zip(leaf_indices);
-            for ((root, _proof), leaf_index) in items {
-                app.database
-                    .insert_pending_identity(leaf_index, &Hash::ZERO, &root)
-                    .await?;
-            }
+            continue;
+        }
```
Dzejkop, line 48: Nice!

src/identity/transaction_manager.rs
```diff
@@ -0,0 +1,545 @@
...
+use crate::database::Database;
+use crate::ethereum::{Ethereum, ReadProvider};
+use crate::identity_tree::{Canonical, Intermediate, TreeVersion, TreeWithNextVersion};
+use crate::prover::Prover;
+use crate::utils::index_packing::pack_indices;
+use crate::utils::retry_tx;
+
+pub type TransactionId = String;
+
+impl From<TransactionId> for crate::ethereum::write::TransactionId {
```
Dzejkop, line 30: `TransactionId` is just a tiny wrapper around `String`, I don't think there's a need for this alias here

src/app.rs
```diff
@@ -27,14 +30,16 @@ use crate::server::error::Error as ServerError;
...
+    pub database:            Arc<Database>,
+    pub identity_manager:    Arc<IdentityManager>,
+    pub transaction_manager: Arc<dyn IdentityTransactionManager + Send + Sync>,
+    tree_state:              OnceLock<TreeState>,
+    pub config:              Config,
 
     pub identity_validator: IdentityValidator,
 }
 
+// impl<T: IdentityTransactionManager> App<T> {
```
Dzejkop, line 42: ?

src/app.rs
```diff
@@ -27,14 +30,16 @@ use crate::server::error::Error as ServerError;
 use crate::utils::retry_tx;
 
 pub struct App {
-    pub database:         Arc<Database>,
-    pub identity_manager: Arc<IdentityManager>,
-    tree_state:           OnceLock<TreeState>,
-    pub config:           Config,
+    pub database:            Arc<Database>,
+    pub identity_manager:    Arc<IdentityManager>,
+    pub transaction_manager: Arc<dyn IdentityTransactionManager + Send + Sync>,
```
Dzejkop, line 35: We could probably make `IdentityTransactionManager: Send + Sync` (`+ 'static` even?) then only leave `Arc<dyn IdentityTransactionManager>` here

src/task_monitor/tasks/delete_identities.rs
```diff
@@ -21,53 +23,60 @@ pub async fn delete_identities(
     let batch_deletion_timeout = chrono::Duration::from_std(app.config.app.batch_deletion_timeout)
         .context("Invalid batch deletion timeout duration")?;
 
+    let mut timer = time::interval(Duration::from_secs(5));
+
     loop {
+        select! {
```
Dzejkop, line 29: No need for this `select`, right?

src/app.rs
piohei: Yeah...sorry. Leftover :(

src/app.rs
piohei: Thanks!

src/identity/transaction_manager.rs
piohei: It was just to not depend on external library with API here. In offchain mode it would be strange to return ethereum::write::TransactionId instead of transaction id from identity module. But I will left decision for others here. ;)

Dzejkop: But it's not an external library we define `TransactionId` in this project

piohei: I left single TransactionId - the one that I added. I removed the one from ethereum crate just to not depend on "ethereum" for off chain mode as it would be odd to me. :)

Dzejkop: Sounds good to me

PR #743, opened 2024-06-12
Refactor code and introduce offchain mode.

piohei:
## Motivation

Chain is not always required to run sequencer. ex. testing, hackathons, etc.

## Solution

Big part of the code had been refactored to keep things separated. Now integration tests are running twice. Once for onchain mode, and once for offchain mode. If some test doesn't make sense for offchain it is not being created.

There is also small fix to task monitor. Now when tasks was failed and we are shutting down process is not aborting. We just not rerun the task.

Also to make it possible to run tests in parallel I have changed how shutdown works. Now it is not a global field but regular struct created at the beginning of the application.

This version still requires semaphore-mtb running. Next step is to remove that requirement for offchain.

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation


src/app.rs
```diff
@@ -50,11 +53,7 @@ impl App {
     /// on the tree state will also error.
     #[instrument(name = "App::new", level = "debug", skip_all)]
     pub async fn new(config: Config) -> anyhow::Result<Arc<Self>> {
-        let ethereum = Ethereum::new(&config);
-        let db = Database::new(&config.database);
-
-        let (ethereum, db) = tokio::try_join!(ethereum, db)?;
```
piohei, line 56: I have removed that sync on etherum and database connnecting. Let me know if this was important part to speed up initialization as I guess connection should be quite fast.

src/config.rs
```diff
@@ -12,15 +12,17 @@ use crate::utils::serde_utils::JsonStrWrapper;
...
+    pub network:       NetworkConfig,
+    pub providers:     ProvidersConfig,
+    pub relayer:       RelayerConfig,
+    pub database:      DatabaseConfig,
+    pub server:        ServerConfig,
     #[serde(default)]
-    pub service:   ServiceConfig,
+    pub service:       ServiceConfig,
+    #[serde(default)]
+    pub offchain_mode: OffchainModeConfig,
```
Dzejkop, line 25: I feel like with off-chain mode we should make network, providers and relayer configs optional

src/app.rs
Dzejkop: Unless there's a noticable difference in tests that's fine by me

src/prover/validator.rs
```diff
@@ -0,0 +1,32 @@
+use anyhow::anyhow;
+
+use crate::config::Config;
+use crate::prover::identity::Identity;
+
+pub struct ProofValidator {
```
Dzejkop, line 6: Not sure I'm seeing the need for this abstraction.

This is equivalent to a function call with 2 args (depth and id commitments).

Dzejkop: Especially since IdentityProcessor impls have a copy of `Config` so they can access tree depth no problem

src/config.rs
piohei: Done

src/prover/validator.rs
piohei: Removed. :)

src/identity/processor.rs
```diff
@@ -20,47 +20,50 @@ use crate::database::query::DatabaseQuery;
...
 
 #[async_trait]
-pub trait IdentityTransactionFinalizer: Send + Sync + 'static {
+pub trait IdentityFinalizer: Send + Sync + 'static {
     async fn finalize_identities(
         &mut self,
         processed_tree: &TreeVersion<Intermediate>,
         mined_tree: &TreeVersion<Canonical>,
     ) -> anyhow::Result<()>;
 }
```
0xKitsune, line 51: Is there a reason that we are creating a trait with corresponding `OffchainFinalizer` and `OnchainFinalizer` structs rather than just implementing a `finalize_roots()` method on the `IdentityProcessor` trait?

With this approach, `identity_processor.finalize_roots()` could return a future or spawned task and we could simply invoke this in the task manager.

piohei: I needed a moment to recall the reason. :D Only `IdentityFinalizer `is mutable while `IdentityProcessor` is not (as stored in app object). Mutability is required by blockchain scanners (like here https://github.com/worldcoin/signup-sequencer/pull/743/files/665a3574a4305633208e1b80eff636ed2f006245#diff-0fa352a4471262fa3eb265e56414ea649dd662b9abde25e8d0e382dd256417baR381). I could go with https://doc.rust-lang.org/book/ch15-05-interior-mutability.html but I personally don't like to use "unsafe" parts. I prefer to let compiler still help with the work. When `IdentityFinalizer` is returned from `IdentityProcessor` then only there the variable (https://github.com/worldcoin/signup-sequencer/pull/743/files/665a3574a4305633208e1b80eff636ed2f006245#diff-5535a609ff4d5028c3f7b86e1fd69d3dd6c5ceeb6bb09938f52a02e8fe55234bR6) is mutable and mutability is "bind" to the owner (here it is `finalize_identities` task).
If you think there is a better solution I'm opened to any other implementation here. :)

0xKitsune: >  I personally don't like to use "unsafe" parts
Interior mutability via `Refcell` or `Mutex` itself not unsafe, even though there is unsafe logic within the implementation. The compiler will still help you if you use a`Mutex` or similar for the mutable components. With that being said, you would still need to make sure you are not acquiring the lock for longer than necessary. 

I think it might be a little cleaner to have the finalization logic directly in the `IdentityProcessor` but I don't have a strong opinion and would defer to @Dzejkop if there is a preference.

Dzejkop: Now that I'm thinking about it we should move `BlockScanner` into `IdentityProcessor` (or somewhere else).

If we create it everytime `finalize_identities` restarts we risk missing blocks that we're not scanning

Dzejkop: And in that case it makes more sense to put it behind a `Mutex`

piohei: Done. Thanks for suggestion! Looks much better now. :)

0xKitsune (approved):

PR #757, opened 2024-06-28
Transactionalize Insert Identities Batch

ewoolsey:
This should solve the error we were getting with serializable transactions failing, specifically when inserting identities. As well I extracted individual transactions out into their own functions for better tx composability.

### DIFF
--- a/src/database/transaction.rs
+++ b/src/database/transaction.rs
@@ -1,10 +1,135 @@
-use sqlx::Executor;
+use sqlx::{Executor, Postgres, Transaction};
+use tokio::sync::Mutex;
 use tracing::instrument;
 
 use crate::database::query::DatabaseQuery;
+use crate::database::types::UnprocessedCommitment;
 use crate::database::{Database, Error};
-use crate::identity_tree::{Hash, ProcessedStatus};
+use crate::identity_tree::{Hash, Latest, ProcessedStatus, TreeVersion, TreeVersionReadOps};
 use crate::utils::retry_tx;
+
+async fn mark_root_as_processed(
+    tx: &mut Transaction<'_, Postgres>,
+    root: &Hash,
+) -> Result<(), Error> {
+    let root_id = tx.get_id_by_root(root).await?;
+
+    let Some(root_id) = root_id else {
+        return Err(Error::MissingRoot { root: *root });
+    };
+
+    let root_id = root_id as i64;
+    // TODO: Can I get rid of line `AND    status <> $2
+    let update_previous_roots = sqlx::query(
+        r#"
+                UPDATE identities
+                SET    status = $2, mined_at = CURRENT_TIMESTAMP
+                WHERE  id <= $1
+                AND    status <> $2
+                AND    status <> $3;
+                "#,
+    )
+    .bind(root_id)
+    .bind(<&str>::from(ProcessedStatus::Processed))
+    .bind(<&str>::from(ProcessedStatus::Mined));
+
+    let update_next_roots = sqlx::query(
+        r#"
+                UPDATE identities
+                SET    status = $2, mined_at = NULL
+                WHERE  id > $1
+                "#,
+    )
+    .bind(root_id)
+    .bind(<&str>::from(ProcessedStatus::Pending));
+
+    tx.execute(update_previous_roots).await?;
+    tx.execute(update_next_roots).await?;
+
+    Ok(())
+}
+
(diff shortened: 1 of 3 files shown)


src/identity/processor.rs
```diff
@@ -143,13 +143,18 @@ impl IdentityProcessor for OnChainIdentityProcessor {
...
-            self.database.delete_batches_after_root(&root_hash).await?;
+            self.database
+                .mark_root_as_processed_and_delete_batches_tx(&root_hash)
+                .await?;
         } else {
             // Db is either empty or we're restarting with a new contract/chain
             // so we should mark everything as pending
-            self.database.mark_all_as_pending().await?;
-            self.database.delete_all_batches().await?;
+            retry_tx!(self.database.pool, tx, {
```
piohei, line 152: It would be nice to put this into transactions.rs file in function to keep it aligned with second if branch. :)

src/database/transaction.rs
```diff
@@ -1,87 +1,180 @@
...
+            tx.remove_unprocessed_identity(&identity.commitment).await?;
+        } else {
+            filtered_identities.push(identity.commitment);
+        }
+    }
+
+    let _guard = pending_insertions_mutex.lock().await;
+
+    let next_db_index = tx.get_next_leaf_index().await?;
+    let next_leaf = latest_tree.next_leaf();
```
piohei, line 106: This line makes me unsure about moving whole function to database module. It is not only modifying database but also latest tree mixing a lot of responsibilities in database module now. What do you think?

src/database/transaction.rs
```diff
@@ -1,87 +1,180 @@
...
+    }
+
+    #[instrument(level = "info", skip_all)]
+    pub async fn insert_identities_batch_tx(
+        &self,
+        latest_tree: &TreeVersion<Latest>,
+        identities: Vec<UnprocessedCommitment>,
+        pending_insertions_mutex: &Mutex<()>,
+    ) -> anyhow::Result<()> {
+        retry_tx!(
```
Dzejkop, line 172: if the underlying function fails we'll still end up in an invalid state because we'll have update the latest_tree, no? So this won't fix the next_leaf_idx desync

src/database/transaction.rs
ewoolsey: Yeah I'm with you. I was internally struggling about where these things belong. On the one hand, it would be nice to have everything that interacts with the database within the database mod. This may be unrealistic and perhaps the concepts of transactions should be allowed to spill over into other code.

src/database/transaction.rs
ewoolsey: Yeah you're right. Somehow I forgot about the interior mutability here. I think something like this with two transactions should be a reasonable solution:

```rust
pub async fn insert_identities_batch(
    database: &Database,
    latest_tree: &TreeVersion<Latest>,
    identities: &[UnprocessedCommitment],
    pending_insertions_mutex: &Mutex<()>,
) -> anyhow::Result<()> {
    let filtered_identities = retry_tx!(database, tx, {
        // Filter out any identities that are already in the `identities` table
        let mut filtered_identities = vec![];
        for identity in identities {
            if tx
                .get_identity_leaf_index(&identity.commitment)
                .await?
                .is_some()
            {
                tracing::warn!(?identity.commitment, "Duplicate identity");
                tx.remove_unprocessed_identity(&identity.commitment).await?;
            } else {
                filtered_identities.push(identity.commitment);
            }
        }
        Result::<_, anyhow::Error>::Ok(filtered_identities)
    })
    .await?;

    let _guard = pending_insertions_mutex.lock().await;

    let data = latest_tree.append_many(&filtered_identities);

    assert_eq!(
        data.len(),
        filtered_identities.len(),
        "Length mismatch when appending identities to tree"
    );

    let next_leaf = latest_tree.next_leaf();

    retry_tx!(database, tx, {
        let next_db_index = tx.get_next_leaf_index().await?;

        assert_eq!(
            next_leaf, next_db_index,
            "Database and tree are out of sync. Next leaf index in tree is: {next_leaf}, in \
             database: {next_db_index}"
        );

        for ((root, _proof, leaf_index), identity) in data.iter().zip(&filtered_identities) {
            tx.insert_pending_identity(*leaf_index, identity, root)
                .await?;

            tx.remove_unprocessed_identity(identity).await?;
        }

        Ok(())
    })
    .await
}
```

What do you think?

Dzejkop (approved):

PR #760, opened 2024-07-09
Send identities in blob or calldata depending on the switch state

cichaczem:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


src/contracts/abi.rs
```diff
@@ -35,6 +35,7 @@ abigen!(
     r#"[
         struct RootInfo { uint256 root; uint128 supersededTimestamp; bool isValid }
         event TreeChanged(uint256 indexed preRoot, uint8 indexed kind, uint256 indexed postRoot)
+        function registerIdentitiesWithBlob(uint256[8] calldata insertionProof, uint256 preRoot, uint32 startIndex, uint256 postRoot) public virtual
```
cichaczem, line 38: We should probably go with method overloading on the contract side

PR #770, opened 2024-09-18
Dzejkop/bump semaphore (and fix CI)

Dzejkop:
Bumps semaphore to newest with oxide branch merged.

Also adds a small CLI tool to do local proof generation and verification

### DIFF
--- a/src/bin/tool.rs
+++ b/src/bin/tool.rs
@@ -1,0 +1,293 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::Context;
+use clap::Parser;
+use ethers::core::rand::{thread_rng, RngCore};
+use semaphore::identity::Identity;
+use semaphore::poseidon_tree::Proof;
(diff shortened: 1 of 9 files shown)


src/bin/tool.rs
```diff
@@ -0,0 +1,293 @@
...
+    identity_file: Option<PathBuf>,
+
+    /// The path to the file that will be used to store the inclusion proof
+    #[clap(long, env)]
+    inclusion_proof_file: Option<PathBuf>,
+
+    /// The path to the file that will be used to store the semaphore proof
+    #[clap(long, env)]
+    semaphore_proof_file: Option<PathBuf>,
+}
```
0xKitsune, line 45: It looks specific args are only used for certian commands. Consider using a similar [approach to forge](https://github.com/foundry-rs/foundry/blob/master/crates/forge/bin/opts.rs#L36-L173) where each command has it's own set of args.

PR #795, opened 2024-10-18
Relax transaction isolation levels + fix insert identities constraints

Dzejkop:


src/app.rs
```diff
@@ -213,76 +209,11 @@ impl App {
...
-    /// Queues a recovery of an identity.
-    ///
-    /// i.e. deletion and reinsertion after a set period of time.
-    ///
-    /// # Errors
-    ///
-    /// Will return `Err` if identity is already queued for deletion, not in the
-    /// tree, or the queue malfunctions.
-    #[instrument(level = "debug", skip(self))]
-    pub async fn recover_identity(
```
Dzejkop, line 243: First step in disabling recovery - deleting the API endpoint

src/task_monitor/tasks/insert_identities.rs
```diff
@@ -42,62 +40,36 @@ pub async fn insert_identities(
...
+            let root = latest_tree.get_root();
 
-    assert_eq!(
-        next_leaf, next_db_index,
-        "Database and tree are out of sync. Next leaf index in tree is: {next_leaf}, in database: \
-         {next_db_index}"
-    );
+            tx.insert_pending_identity(leaf_idx, identity, &root, &pre_root)
+                .await
+                .expect("Failed to insert identity - tree will be out of sync");
```
Dzejkop, line 61: The most critical change in this PR is this and the other expect on line 70.

These are necessary to keep the sequencer state in sync with the db

src/app.rs
```diff
@@ -213,76 +209,11 @@ impl App {
             return Err(ServerError::IdentityAlreadyDeleted);
         }
 
-        // Check if the id is already queued for deletion
-        if tx.identity_is_queued_for_deletion(commitment).await? {
```
Dzejkop, line 217: the `insert_new_deletion` method has been changed to be idempotent. Therefore there's no longer a need to check if we have already accepted a deletion

0xKitsune (approved):

0xKitsune (approved):

src/task_monitor/tasks/insert_identities.rs
0xOsiris: I guess my only question is will the `identity` be completely lost here if this fails? And therein all unprocessed identities prior? Same below with `tx.commit()` failing. Just wondering what the flow would be after restart

0xOsiris (approved):

PR #797, opened 2024-10-21
Remove recovery related code

Dzejkop:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


0xOsiris (approved):
This LGTM, only concern is this endpoint is stilled hooked into production in app/maybe other places. Have we notified teams this endpoint will not exist anymore?

ewoolsey (approved):
That's a lot of removed code. Feels good.

Dzejkop:
> This LGTM, only concern is this endpoint is stilled hooked into production in app/maybe other places. Have we notified teams this endpoint will not exist anymore?
yep, and we've not been receiving any traffic to this endpoint in over a month

PR #805, opened 2024-10-25
Graceful Shutdown

ewoolsey:
## Motivation

Currently we've been running with `panic="abort"` which is known to cause problems with our mmap caching system.

## Solution

I've overhauled the shutdown system that we've been using to 
- make it more ergonomic
- make it more functional
- allow for graceful shutdowns through `std::process::exit`

There are now 2 ways we can shutdown long running tasks
1. - Cancel the future they're running on.
        - Allows tasks to continue for a short delay period until they hit an `await` point and stall.
        - This should work for 99% of cases
2. - Provide the task with it's own `Shutdown` instance and allow it some short period (30s default) to clean itself up.

Since all panics used to abort, It's possible we may have been relying on that behaviour in some places I haven't noticed. Instead now we should prefer issuing a `shutdown`.

For this reason, I've installed a `PanicHandler` layer in the server to catch panics then issue a global shutdown. I'm not actually sure this is what we want so I'm happy to remove it if that's the case.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation

### DIFF
--- a/src/shutdown.rs
+++ b/src/shutdown.rs
@@ -63,7 +63,7 @@ impl Shutdown {
 
     /// Wait for the shutdown to begin.
     ///
-    /// If the caller down not already own a `Receiver`
+    /// If the caller does not already own a `Receiver`
     /// then the process may abort immediately when this returns.
     pub async fn await_shutdown_begin(&self) {
         let _ = self.await_shutdown_begin_with_handle().await;
--- a/src/shutdown.rs
+++ b/src/shutdown.rs
@@ -1,66 +1,116 @@
-use std::future::Future;
-use std::sync::Arc;
+use std::time::Duration;
 
 use eyre::Result;
+use tokio::select;
 use tokio::sync::watch::{self, Receiver, Sender};
-use tracing::info;
+use tracing::{error, info};
 
+/// A watch channel used to syncronize tasks to shutdown.
+#[derive(Clone)]
 pub struct Shutdown {
     sender: Sender<bool>,
-    receiver: Receiver<bool>,
 }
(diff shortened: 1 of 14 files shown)


cichaczem (approved):

cichaczem:
LGTM

I'm looping in @jazzbee, with whom we talked about gracefully shot down some time ago.

src/shutdown.rs
```diff
@@ -1,66 +1,116 @@
...
 
-impl Default for Shutdown {
-    fn default() -> Self {
-        Self::new()
+    /// Wait for the shutdown to begin.
+    ///
+    /// If the caller down not already own a `Receiver`
+    /// then the process may abort immediately when this returns.
+    pub async fn await_shutdown_begin(&self) {
+        let _ = self.await_shutdown_begin_with_handle().await;
```
Dzejkop, line 69:
```suggestion
        let _ = self.await_shutdown_begin_with_handle().await.await;
```
?

As it stands the type of `_` is `Receiver<bool>`

Dzejkop (approved):
LGTM! Thanks!

src/shutdown.rs
ewoolsey: I'm not sure what you're suggesting I don't think? This looks as intended to me. `Receiver<boo>` is `must_use` so I have to access it like this to throw it away. You also cannot `await` a `Receiver`.

ewoolsey:
@Dzejkop My most recent commit implements a shutdown on panic for the task manager as well as some other cleanup/simplification.

PR #815, opened 2024-11-06
Fix tree restore on startup.

piohei:
## Motivation

Fixing how tree is being restored.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/identity_tree/initializer.rs
+++ b/src/identity_tree/initializer.rs
@@ -47,27 +47,39 @@ impl TreeInitializer {
             .await?;
 
         let timer = Instant::now();
-        let mut tree_state = self.restore_or_initialize_tree(initial_root_hash).await?;
+        info!("Tree state initialization started");
+        let tree_state = self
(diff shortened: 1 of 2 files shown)


src/identity_tree/initializer.rs
```diff
@@ -78,7 +90,7 @@ impl TreeInitializer {
 
         let mined_items = dedup_tree_updates(mined_items);
 
-        if !self.config.force_cache_purge {
+        if !force_cache_purge {
```
Dzejkop, line 93: this is still the same value? We can just use `self.config.force_cache_purge` no?

Dzejkop (approved):

src/identity_tree/initializer.rs
piohei: For second call here https://github.com/worldcoin/signup-sequencer/pull/815/files#diff-564578c6f713fb19e94efd82c09256f2b2f8c831854818c48a8b2b260524a30dR69 it is set to static true to force cache purge.

PR #827, opened 2024-11-25
Fix batch creation.

piohei:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/task_monitor/tasks/create_batches.rs
+++ b/src/task_monitor/tasks/create_batches.rs
@@ -1,11 +1,11 @@
-use std::sync::Arc;
-
 use anyhow::Context;
 use chrono::{DateTime, Utc};
 use ethers::prelude::U256;
 use ruint::Uint;
 use semaphore::merkle_tree::Proof;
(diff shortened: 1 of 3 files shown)


src/task_monitor/tasks/create_batches.rs
```diff
@@ -39,7 +39,7 @@ pub async fn create_batches(
 
     // We start a timer and force it to perform one initial tick to avoid an
     // immediate trigger.
-    let mut timer = time::interval(app.config.app.batch_insertion_timeout);
+    let mut timer = time::interval(Duration::from_secs(5));
```
cichaczem, line 42: Why change from a config value to a constant one?

cichaczem (approved):

src/task_monitor/tasks/create_batches.rs
piohei: We use 5 seconds constant in other tasks and insertion timeout is very big. And when restarted it was waiting batch_insertion timeout before doing a batch while there was enough identities already waiting. :)

PR #831, opened 2024-12-02
Fix finalization for offchain mode.

piohei:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/identity/processor.rs
+++ b/src/identity/processor.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::sync::{Arc, Mutex};
 
 use anyhow::anyhow;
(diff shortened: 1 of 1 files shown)


src/identity/processor.rs
```diff
@@ -529,19 +529,31 @@ impl IdentityProcessor for OffChainIdentityProcessor {
...
                 .begin_tx(IsolationLevel::ReadCommitted)
                 .await?;
 
+            let root_state = tx.get_root_state(&batch.next_root).await?;
+            if root_state.is_none() {
+                // If root is not in identities table we can't mark it as processed or mined.
+                // It happens sometimes as we do not have atomic operation for database and tree
+                // insertion.
+                // TODO: check if this is still possible after HA being done
+                tx.commit().await?;
```
Dzejkop, line 553: wouldn't rollback be more fitting here? Since we intend for nothing to happen?

piohei: I would rather leave commit here as there was no error at all.

Dzejkop (approved):

PR #842, opened 2025-01-07
Sync in-memory trees using DB (DB is source of truth now).

piohei:
## Motivation

For Highly Available sequencer we will use database as source of truth and synchronization point.

## Solution

Changes:
* new task introduced to sync with database
* for tree now there is an option to move forward and backward in history
* added e2e tests for offchain mode
* processing insertions/deletions is now consolidated into single task to give deletions precedence over insertions and also simplify logic with mutexes
* identity tree refactored
* for offchain mode transactions are being finalized based on database state instead of internal in-memory queue

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation


e2e_tests/scenarios/tests/common/docker_compose.rs
```diff
@@ -266,19 +305,17 @@ fn run_cmd(cwd: &str, envs: HashMap<String, String>, cmd_str: String) -> anyhow:
...
-        .unwrap()
-        .split(':')
-        .last()
-        .unwrap()
-        .parse::<u32>()
-        .unwrap()
+    let port = parts.last().and_then(|v| v.split(':').last());
+
+    match port {
+        Some(port) => port.parse::<u32>().map_err(|err| anyhow!(err)),
```
Dzejkop, line 318: For tests I often find that `.expect` or `.unwrap` if often better because you get line number information on where the error occured.

Which anyhow you'll get a vague "Expected number found 'a'" which makes it more difficult to nail down.

I would suggest adding `.context` if sticking with anyhow

src/ethereum/write_provider/openzeppelin.rs
```diff
@@ -135,6 +135,7 @@ impl OzRelay {
         &self,
         mut tx: TypedTransaction,
         only_once: bool,
+        _tx_id: Option<String>,
```
Dzejkop, line 138: worth noting - we don't use OZ anymore anywhere AFAIK, we should remove this module in a future PR

src/identity_tree/db_sync.rs
```diff
@@ -0,0 +1,216 @@
...
+    let latest_mined_tree_update = tx
+        .get_latest_tree_update_by_statuses(vec![ProcessedStatus::Mined])
+        .await?;
+
+    // First check if mined tree needs to be rolled back. If so then we must
+    // panic to quit to rebuild the tree on startup. This is a time-consuming
+    // operation.
+    if let Some(ref mined_tree_update) = latest_mined_tree_update {
+        assert!(
+            mined_tree_update.sequence_id >= mined_tree.get_last_sequence_id(),
```
Dzejkop, line 31: shouldn't this condition be flipped? i.e. our last in-memory id is larger than the one in the db?

src/identity_tree/db_sync.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    let latest_batch = tx.get_latest_batch().await?;
+    let latest_batching_tree_update = if let Some(latest_batch) = latest_batch {
+        tx.get_tree_update_by_root(&latest_batch.next_root).await?
+    } else {
+        latest_processed_tree_update.clone()
+    };
+
+    // And then update trees
+    update_latest_tree(tx, latest_tree, &latest_pending_tree_update, || {
```
Dzejkop, line 60: very nice! Clever approach

docs/images/flows-Identity flow.png
Dzejkop, line 1: could you export this one in higher resolution? The text is hard to read

src/identity_tree/db_sync.rs
piohei: It is correct. This is assert so you pass condition that fails on being false. Other words, I'm ensuring here that record taken from database has sequence id greater or equal than sequence id of in-memory tree. But thanks for double checking! :)

e2e_tests/scenarios/tests/common/docker_compose.rs
piohei: Good point - I have removed all error mapping and replaced it with parse.

src/identity/processor.rs
```diff
@@ -409,37 +407,49 @@ impl OnChainIdentityProcessor {
...
-                .begin_tx(IsolationLevel::ReadCommitted)
-                .await?;
-            tx.mark_root_as_processed(&post_root.into()).await?;
-            tx.commit().await?;
+            retry_tx!(self.database.pool, tx, {
+                // With current flow it is required to mark root as processed first as this is
+                // how required mined_at field is set, We set proper state only if not set
+                // previously.
+                let root_state = tx.get_root_state(&post_root).await?;
+                match root_state {
```
Dzejkop, line 438: this is a little confusing to read? Maybe this should be an `if not ...` ?

src/task_monitor/tasks/modify_tree.rs
```diff
@@ -0,0 +1,231 @@
...
+    info!("Starting modify tree task.");
+
+    let batch_deletion_timeout = chrono::Duration::from_std(app.config.app.batch_deletion_timeout)
+        .context("Invalid batch deletion timeout duration")?;
+    let min_batch_deletion_size = app.config.app.min_batch_deletion_size;
+
+    let mut timer = time::interval(Duration::from_secs(5));
+
+    loop {
+        // We wait either for a timer tick or a full batch
```
Dzejkop, line 37: comment incorrect - it's timer or db event?

src/task_monitor/mod.rs
```diff
@@ -62,10 +64,17 @@ impl TaskMonitor {
...
-        base_wake_up_notify.notify_one();
+        let base_next_batch_notify = Arc::new(Notify::new());
+        // Immediately notify, so we can start processing if we have pending operations
+        base_next_batch_notify.notify_one();
+
+        let base_sync_tree_notify = Arc::new(Notify::new());
+        // Immediately notify, so we can start processing if we have pending operations
+        base_sync_tree_notify.notify_one();
+
+        let (base_tree_synced_tx, base_tree_synced_rx) = watch::channel(false);
```
Dzejkop, line 75: we only ever use `.changed()` on this watch channel. How is that different from a `Notify`?

src/identity/processor.rs
```diff
@@ -409,37 +407,49 @@ impl OnChainIdentityProcessor {
...
                 continue;
             }
 
-            let mut tx = self
-                .database
-                .begin_tx(IsolationLevel::ReadCommitted)
-                .await?;
-            tx.mark_root_as_processed(&post_root.into()).await?;
-            tx.commit().await?;
+            retry_tx!(self.database.pool, tx, {
```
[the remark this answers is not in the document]
piohei: I'm retrying just to not wait for next log. The error here might be due to a lot of reasons I guess. I would say it is better to have it here.

src/task_monitor/mod.rs
piohei: I changed `Notify` to `Channel` because I need to be able to listen for the event in multiple places. With `Notify` only single watcher will be notified about the event.

Dzejkop: I think notify_waiters notifies multiple places https://docs.rs/tokio/1.43.0/tokio/sync/struct.Notify.html#method.notify_waiters

src/ethereum/write_provider/openzeppelin.rs
Dzejkop: Awesome, thanks!

src/task_monitor/mod.rs
Dzejkop: After further thinking watch seems like a better tool here. `.notify_waiters` would notify every sleeping task, but we'd lose events which happened while a task was busy.

But we could change the type of the channel to `()`

src/task_monitor/tasks/modify_tree.rs
```diff
@@ -0,0 +1,231 @@
...
+    // they are continuous ending with last leaf index
+    deletions.sort_by(|d1, d2| d1.leaf_index.cmp(&d2.leaf_index));
+
+    if let Some(last_leaf_index) = tree_state.latest_tree().next_leaf().checked_sub(1) {
+        let indices_are_continuous = deletions
+            .windows(2)
+            .all(|w| w[1].leaf_index == w[0].leaf_index + 1);
+
+        if indices_are_continuous && deletions.last().unwrap().leaf_index == last_leaf_index {
+            warn!(
```
Dzejkop, line 150: I'm a little worried that under certain conditions we'd see this warning many times in 5 second intervals (or more frequent).

I'm wondering if it makes sense to bump the deletion timestamp here to delay it further. Or keep an in memory timer just for this condition

src/identity/processor.rs
piohei: changed as discussed

src/task_monitor/tasks/modify_tree.rs
Dzejkop: sounds good

e2e_tests/scenarios/tests/common/docker_compose.rs
piohei: I can't change it. Using `.unwrap()` is generating panic while I need `Result` here to retry the operation few times.

Dzejkop: ok, makes sense

src/ethereum/write_provider/tx_sitter.rs
```diff
@@ -4,8 +4,9 @@ use anyhow::Context;
 use async_trait::async_trait;
 use ethers::types::transaction::eip2718::TypedTransaction;
 use ethers::types::U256;
+use reqwest::StatusCode;
 use tx_sitter_client::data::{SendTxRequest, TransactionPriority, TxStatus};
-use tx_sitter_client::TxSitterClient;
+use tx_sitter_client::{HttpError, TxSitterClient};
```
Dzejkop, line 9: Compliation error:
```
   Compiling signup-sequencer v2.0.0 (/Users/jakub.trad/Projects/Worldcoin/signup-sequencer)
error[E0432]: unresolved import `tx_sitter_client::HttpError`
 --> src/ethereum/write_provider/tx_sitter.rs:9:24
  |
9 | use tx_sitter_client::{HttpError, TxSitterClient};
  |                        ^^^^^^^^^ no `HttpError` in the root

For more information about this error, try `rustc --explain E0432`.
```

piohei: It's here https://github.com/worldcoin/signup-sequencer/blob/2dc572d27a8728dafaee9385f23b5d0bcf58fb59/crates/tx-sitter-client/src/lib.rs#L18
Interesting 🤔 Working fine on my machine (acutally on two of them and here in github actions when running checks).

tests/tree_restore_with_root_back_to_init.rs
```diff
@@ -109,9 +112,22 @@ async fn tree_restore_with_root_back_to_init(offchain_mode_enabled: bool) -> any
     )
     .await;
 
-    tokio::time::sleep(Duration::from_secs(2)).await;
+    tokio::time::sleep(Duration::from_secs(5)).await;
+
+    // Await for tree to be mined in app
+    let tree_state = {
```
Dzejkop, line 118: we reuse this block across multiple tests - can we extract it to a function?

src/identity_tree/db_sync.rs
Dzejkop: gotcha, makes sense!

src/ethereum/write_provider/tx_sitter.rs
Dzejkop: nevermind, was some weird caching issue on my machine

Dzejkop (approved):

tests/tree_restore_with_root_back_to_init.rs
piohei: Done :)

PR #878, opened 2025-05-14
Piohei/impl api v2

piohei:
## Motivation

New API version should be easier to use and maintain.

## PR Checklist

- [ X ] Added Tests
- [ X ] Added Documentation


src/app/mod.rs
```diff
@@ -356,6 +482,78 @@ impl App {
...
+
+        // If tree change was applied to latest tree we look for inclusion proof
+        // todo(piotrh): is it ok to use latest tree here?
+        let (leaf, root, proof) = self
+            .tree_state()?
+            .get_latest_tree()
+            .get_leaf_and_proof(item.leaf_index);
+
+        if leaf != commitment {
+            return Err(InclusionProofV2Error::InvalidCommitment);
```
Dzejkop, line 549: This should be an internal server error of some sort. When would this ever happen?

src/app/mod.rs
```diff
@@ -457,4 +661,107 @@ impl App {
...
+                } else {
+                    error!("Root state does not have mined_at set while have mined status. This should never happen. Considering root age too old.");
+                    return Err(anyhow::Error::msg(
+                        "Unexpected error ocurred while checking root age.",
+                    ));
+                }
+            }
+        };
+
+        warn!("Root age: {root_age:?}");
```
Dzejkop, line 763:
```suggestion
```

src/app/mod.rs
```diff
@@ -457,4 +661,107 @@ impl App {
...
+        &self,
+        max_root_age: Duration,
+        root_state: &RootItem,
+    ) -> anyhow::Result<bool> {
+        let tree_state = self.tree_state()?;
+        let latest_root = tree_state.get_latest_tree().get_root();
+        let processed_root = tree_state.get_processed_tree().get_root();
+        let mined_root = tree_state.get_mined_tree().get_root();
+
+        info!("Validating age max_root_age: {max_root_age:?}");
```
Dzejkop, line 723:
```suggestion
        info!(?max_root_age, "Validating age max_root_age");
```

or something like this, I forgot the exact syntax

src/app/mod.rs
```diff
@@ -457,4 +661,107 @@ impl App {
...
+
+        let root = root_state.root;
+        match root_state.status {
+            // Pending status implies the batching or latest tree, but batching tree is only
+            // for internal processing purposes
+            ProcessedStatus::Pending if latest_root == root => {
+                warn!("Root matches current pending tree root - considering valid");
```
Dzejkop, line 730:
```suggestion
```

src/app/mod.rs
```diff
@@ -457,4 +661,107 @@ impl App {
...
+                return Ok(true);
+            }
+            // Processed status implies the processed tree
+            ProcessedStatus::Processed if processed_root == root => {
+                warn!("Root matches current processed tree root - considering valid");
+                return Ok(true);
+            }
+            // Processed status implies the mined tree
+            ProcessedStatus::Mined if mined_root == root => {
+                warn!("Root matches current mined tree root - considering valid");
```
Dzejkop, line 740:
```suggestion
```

src/app/error.rs
```diff
@@ -0,0 +1,75 @@
...
+    #[error("provided root is too old")]
+    RootTooOld,
+    #[error("cannot decompress provided proof")]
+    DecompressingProofError,
+    #[error("prover error")]
+    ProverError,
+    #[error(transparent)]
+    Database(#[from] database::Error),
+    #[error(transparent)]
+    AnyhowError(#[from] anyhow::Error),
```
Dzejkop, line 74: A friend enlightened me recently to a cool pattern, where instead of doing this, we go for:

```suggestion
    RootValidityError(anyhow::Error),
```

which makes it a little more difficult to use but it makes the choice of the error explicit. So at the error propagation site we have to do something like

```rust
let is_root_valid = self.validate_root_age(root_age)
    .map_err(VerifySemaphoreProofV2Error::RootValidityError)?;
```

src/server/api_v2/custom_middleware/remove_auth_layer.rs
```diff
@@ -0,0 +1,13 @@
+use axum::extract::Request;
+use axum::http::header::AUTHORIZATION;
+use axum::http::StatusCode;
+use axum::middleware::Next;
+use axum::response::Response;
+
+pub async fn middleware(mut request: Request, next: Next) -> Result<Response, StatusCode> {
```
Dzejkop, line 7: is this the exact same code as in the api v1? Can we share common middleware?

src/server/api_v2/mod.rs
```diff
@@ -0,0 +1,1298 @@
...
+        provers_urls = "[]"
+
+        [tree]
+        dense_tree_prefix_depth = 8
+
+        [database]
+        database = "postgres://user:password@localhost:5432/database"
+
+        [server]
+        address = "0.0.0.0:3001"
```
Dzejkop, line 320: does this mean all tests using this function spin up a server at port 3001? We should change this to `127.0.0.1` to avoid the annoying MacOs popups (also it just makes sense) and consider choosing a random port

src/app/mod.rs
piohei: Do we even need this log as we are removing all the others? I'm gonna remove it.

src/server/api_v2/mod.rs
piohei: We do not spin up server in testing here (but good point with config here!).

src/server/api_v2/mod.rs
```diff
@@ -306,33 +307,45 @@ mod test {
...
+                batch_insertion_timeout: default::batch_insertion_timeout(),
+                batch_deletion_timeout: default::batch_deletion_timeout(),
+                min_batch_deletion_size: default::min_batch_deletion_size(),
+                scanning_window_size: default::scanning_window_size(),
+                scanning_chain_head_offset: default::scanning_chain_head_offset(),
+                time_between_scans: default::time_between_scans(),
+                monitored_txs_capacity: default::monitored_txs_capacity(),
+                shutdown_timeout: default::shutdown_timeout(),
+                shutdown_delay: default::shutdown_delay(),
+            },
```
Dzejkop, line 324: Could we do something like
```rust
Config {
    tree: TreeConfig {
        dense_tree_prefix_depth: 8,
        ..Default::default()
    },
    ..Default::default()
}
```

piohei: We don't have default defined for config. :/

Dzejkop (approved):

PR #884, opened 2025-06-24
Piohei/fix batches creation

piohei:
## Motivation

Insertion batch is not being instantly created when not full and there is deletion batch ahead. It should then not wait for anyting. Now it awaits for insertion timeout to be created.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation


0xForerunner (approved):
This looks good to me @piohei.

PR #886, opened 2025-07-07
Add mutex guard around tree state.

piohei:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation


piohei:
> I personally don't think the right architecture is to have long held mutexes to prevent background state modification. I think it's likely better to quickly acquire the mutex, fetch all the state we need, then finish the task with that snapshot of the state.

I know - I'm thinking same in general. Here the create_batches operation should be quick. It is separated from processing the batch. The sync state is crucial operation which shuold be atomic in general. In otther places, as you might noticed, the lock is just for a moment to get data we need.

I will check perf on staging to see if this is problematic in any way.

PR #910, opened 2026-01-21
Authentication layer

Dzejkop:
This PR adds an authentication layer which supports both basic auth and JWT auth based on a known set of trusted public keys.

### DIFF
--- a/src/utils/auth.rs
+++ b/src/utils/auth.rs
@@ -96,10 +96,10 @@ impl AuthValidator {
             Some(token) => {
                 // Token present - validate it
                 match self.jwt_validator.validate(token) {
-                    Ok(key_name) => {
+                    Ok(claims) => {
                         tracing::info!(
                             basic_user = %basic_username,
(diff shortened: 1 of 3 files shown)


src/server/api_v1/custom_middleware/auth_layer.rs
```diff
@@ -0,0 +1,24 @@
...
+) -> Response {
+    match validator.validate(&request) {
+        AuthResult::Allowed => next.run(request).await,
+        AuthResult::AllowedWithWarning(msg) => {
+            tracing::warn!("{}", msg);
+            next.run(request).await
+        }
+        AuthResult::Denied(msg) => {
+            tracing::warn!("Authentication failed: {}", msg);
+            (StatusCode::UNAUTHORIZED, "Unauthorized").into_response()
```
alessandromazza98: nit: you may use the same error message you use in `api_v2/custom_middleware/auth_layer` to be more consistent

PR #917, opened 2026-02-09
Piohei/allow readding removed ids

piohei:
## Motivation

This PR introduces two changes:
* New API v3 that allows to re-add previously deleted commitment
* New inclusion-proof endpoint in API v3 that allows to specify which tree state use to get proof

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/tests/add_remove_readd_identity.rs
+++ b/tests/add_remove_readd_identity.rs
@@ -1,0 +1,284 @@
+#![allow(clippy::needless_range_loop)]
+
+mod common;
+
+use common::prelude::*;
+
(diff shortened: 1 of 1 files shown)


src/database/methods.rs
```diff
@@ -235,6 +235,30 @@ pub trait DbMethods<'c>: Acquire<'c, Database = Postgres> + Sized {
         .await?)
     }
 
+    #[instrument(skip(self), level = "debug")]
+    async fn get_tree_item_by_statuses(
```
Dzejkop: can we add a test?

PR #936, opened 2026-04-08
feat: upgrade authorized_keys to support multiple PEM entries with optional expiry.

piohei:
- `authorized_keys` type changes from `HashMap<String, String>` to `HashMap<String, Vec<AuthorizedKey>>`, where `AuthorizedKey` carries a PEM string and an optional `expires_at` timestamp
- JWT validation skips keys whose `expires_at` is in the past, enabling zero-downtime key rotation by listing old and new keys side-by-side
- Env var renamed to `SEQ__SERVER__AUTHORIZED_KEYS_JSON` and accepts a JSON map (`{"name":[{"pem":"...","expires_at":"..."}]}`); the old `authorized_keys` TOML key is kept as a serde alias for compatibility
- Custom serde deserializer handles both TOML sequences and JSON-encoded strings from env vars, including the full inline-map syntax

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -43,8 +43,7 @@ mod authorized_keys_serde {
             }
 
             fn visit_str<E: serde::de::Error>(self, v: &str) -> Result<Self::Value, E> {
-                serde_json::from_str::<HashMap<String, Vec<AuthorizedKey>>>(v)
-                    .map_err(E::custom)
+                serde_json::from_str::<HashMap<String, Vec<AuthorizedKey>>>(v).map_err(E::custom)
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/utils/jwt.rs
```diff
@@ -65,11 +80,17 @@ impl JwtValidator {
...
 
-        for (name, key) in &self.keys {
-            if let Ok(token_data) = decode::<Claims>(token, key, &validation) {
-                // Signature valid - now check sub matches key name
-                if token_data.claims.sub == *name {
-                    return Ok(token_data.claims);
+        let now = Utc::now();
+        for (name, keys) in &self.keys {
+            for loaded in keys {
+                if loaded.expires_at.is_some_and(|exp| exp <= now) {
```
Dzejkop: Idea - maybe we could validate the key and only check expiry after so we could log a warning with a key name/id?

src/config.rs
```diff
@@ -253,9 +334,14 @@ pub struct ServerConfig {
...
 
-    /// Named authorized keys for JWT authentication: key_name -> PEM public key
-    #[serde(default)]
-    pub authorized_keys: HashMap<String, String>,
+    /// Named authorized keys for JWT authentication: key_name -> list of authorized keys
+    #[serde(
+        rename = "authorized_keys_json",
+        alias = "authorized_keys",
+        default,
+        deserialize_with = "authorized_keys_serde::deserialize"
```
Dzejkop, line 342: there's also a `with` option 


```suggestion
        with = "authorized_keys_serde"
```

which should point to both the deserialization & serialization methods

Dzejkop (approved):

src/utils/jwt.rs
piohei: good idea - added!

PR #955, opened 2026-07-29
feat: abandon invalid batches

0xForerunner:


src/task_monitor/tasks/process_batches.rs
0xOsiris, line 70: Shouldn't the below be happening prior to this?

0xOsiris:
@codex review

0xOsiris (approved):

